PR #2, opened 2026-07-16
feat(gate): first-class fail-closed "strict" gate (darnlink-strict) + docs + dogfood

txemi:
## Why

darnlink's published quality gate only keeps the robust links you **already have** from breaking. It says nothing about **plain** relative links that were never anchored — so a fresh, un-anchored link sails through the gate and breaks on the next refactor.

The fail-closed check (`darnlink --robustify` in dry-run) already existed and exits non-zero when an anchorable plain link is un-anchored ([`cli.py:97`](https://github.com/txemi/darnlink/blob/main/src/darnlink/cli.py#L97)) — but it was **undocumented**, so downstream repos reinvented it as bespoke scripts. This makes it first-class.

## What

- **`.pre-commit-hooks.yaml`**: new `darnlink-strict` hook id → `darnlink --robustify` (dry-run, no writes).
- **README**: new *"Stricter: require every link to be robust (fail-closed)"* section — what it catches beyond the default, the generated-file escape hatch (`<!-- darnlink-ignore-file -->` / `--exclude`), and adoption steps.
- **Dogfood**: anchored the one plain link in `CHANGELOG.md` (→ `FORMAT.md`) and gated darnlink's own repo with the strict check in CI (`ci.yml`) and `tools/check.sh`.
- **CHANGELOG**: `Unreleased` entry.

Fail-closed semantics: it fails until every link that *can* be robust *is* robust. Links whose target has no `uuid` (external files, generated exports, mirrors) are left alone.

## Notes for review
- No new flags or code paths — the mechanism (`--robustify` dry-run) is unchanged; this exposes + documents + dogfoods it.
- The README's `darnlink-strict` example flags that the hook ships in the **next release** (Unreleased), so pinning `rev: v0.1.1` + `id: darnlink-strict` won't silently 404.
- Local gate + CI green: 52 tests pass, repair check clean, strict check clean.

## Test plan
- [x] `uv run pytest -q` → 52 passed
- [x] `darnlink .` (repair) → clean
- [x] `darnlink . --robustify` (strict) → clean after dogfood anchor

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,6 +27,8 @@ jobs:
         run: uv run pytest -q
       - name: darnlink self-check (links must be clean)
         run: uv run darnlink .
+      - name: darnlink strict self-check (fail-closed — every anchorable link must be robust)
+        run: uv run darnlink . --robustify
 
   # Exercises the commands users actually run (packaging + entry point + the literal README
   # one-liner) on Linux AND Windows — so the copy-paste one-liner is proven on both.
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -22,9 +22,10 @@
 - id: darnlink-strict
   name: darnlink strict (fail-closed — require every anchorable link to be robust)
   description: >-
-    Fail (dry-run, no writes) if any plain Markdown link whose target already has a uuid is left
-    un-anchored. Stricter than `darnlink`: it does not just keep existing robust links working, it
-    requires links to be robust in the first place. Exempt generated files with a
+    Fail (dry-run, no writes) if any plain Markdown link to an anchorable target (a local file with
+    frontmatter) is left un-anchored. Stricter than `darnlink`: it does not just keep existing robust
+    links working, it requires links to be robust in the first place. Targets that can't take a uuid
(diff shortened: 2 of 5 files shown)


txemi:
Thanks @Copilot — all 6 suggestions were valid and are addressed in 8fe66d4:

1. **"target has no uuid" wording** (README ×2, CHANGELOG, `.pre-commit-hooks.yaml`) — corrected. The exemption now reads: targets that *can't take a uuid* — non-local, deny-listed, or **without frontmatter** (unless `--create-frontmatter`). `--robustify` does create/reuse a uuid for targets that already have frontmatter, so the old phrasing was wrong.
2. **`rev: v0.1.1` + `id: darnlink-strict`** — the strict example now pins `rev: v0.2.0` (first tag that will ship the hook), with a note to pin a `main` SHA until it's tagged. The other example (which uses only `darnlink`/`darnlink-repair`, present at v0.1.1) is unchanged.
3. **`[Unreleased]` reference link** — added `[Unreleased]: …/compare/v0.1.1...HEAD` at the bottom.

Re-requesting review.

txemi:
**Verdict: ✅ merge.** Copilot re-review (16:48) reported *"reviewed 5 out of 5 changed files and generated no new comments"* — all 6 findings from the first pass addressed. CI green (test + smoke, Linux + Windows, all Python versions). Merging (squash).

PR #4, opened 2026-07-17
docs: full option reference, in-file vocabulary, and reviewer/gate guidance

txemi:
Docs-only follow-up to #3. No code change; `git diff --stat` is README / FORMAT.md / CONTRIBUTING.md.
Fills gaps found while actually using the tool this week.

## README
- **`## All options`** — the full flag list was scattered across prose, and **`--json` was documented
  nowhere**, even though it's the contract gates parse. Adds a reference table, the stable `--json`
  shape, and the list of `kind`s — with the note to count the kinds you care about rather than
  `len(findings)` (the non-actionable kinds are emitted precisely so nothing is skipped silently).
- **Strict-gate scope note** — every published hook runs over the whole tree (`pass_filenames:
  false`; darnlink takes a directory, not a file list) and strict is fail-closed, so a plain link
  *someone else* left in a file you never touched blocks *your* commit. Fine solo; with several
  contributors it deadlocks. Documents the practical split: `darnlink-strict` in CI (the real wall),
  the plain `darnlink` hook locally.

## FORMAT.md §5
- Adds the one in-file construct that was missing here — the `<!-- NAME-start --> … <!-- NAME-end -->`
  region (it lived only in the README, as if it were a flag) — plus a summary table of the **whole**
  in-file vocabulary, so §5 is now the single place that answers "what can a file say to darnlink?".
- The `## 5. Opting a file out` heading is kept verbatim, so the anchor other docs link to
  (`#5-opting-a-file-out`) still resolves.

## CONTRIBUTING.md
- Documents that Copilot review is enabled and how to work with it — including the two things that
  cost time this week: it reviews **once, on open** (a re-review must be *requested*), and the REST
  `requested_reviewers` endpoint **silently ignores the bot**, so you need GraphQL `botIds`.

## Checks
Docs only, but darnlink dogfoods its own links: the strict gate flagged the two new FORMAT.md links I
added (now anchored), and `darnlink .` + `darnlink . --robustify` + the 58 tests stay green.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -43,6 +43,28 @@ CI (`.github/workflows/ci.yml`) runs on every PR; please make sure it's green. A
 - The robust-link format is specified in [FORMAT.md](FORMAT.md) <!-- uuid: 9052d864-2a45-4ed4-8725-d8a394e7a7ef --> — it's tool-agnostic, so changes to
   the format are a bigger deal than changes to the tool; discuss them first.
 
+### Automated review
+
+This repo has **GitHub Copilot code review** enabled. When you open a PR, Copilot reviews it
+automatically and leaves inline comments. Please address them before asking a human to merge: fix the
+valid ones, and reply to any you're dismissing with the reason. It has a good hit rate here — on a
+recent PR all four of its findings were real (two correctness, two performance).
+
+Two gotchas worth knowing:
+
+- **It reviews once, on open.** Pushing a fix does **not** trigger a re-review — you have to request
+  one. In the GitHub UI: the *Reviewers* gear → re-request Copilot.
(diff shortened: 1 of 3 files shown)


txemi:
## Verdict: ready to merge

Docs-only (README / FORMAT.md / CONTRIBUTING.md — no code). CI 20/20 green, MERGEABLE. Copilot: 1 round, **no findings** — its review is an approving summary. No second round requested: on #3 that was justified because Copilot was surfacing real defects; here there's no code to find any in, so re-requesting would be ceremony.

darnlink dogfoods its own links: the strict gate caught the two new FORMAT.md links this PR added (now anchored), and `darnlink .` + `darnlink . --robustify` + the 58 tests stay green.

Merging.

PR #6, opened 2026-07-17
feat(check): `darnlink check` — report-only gate, both axes, distinguishable exit codes

txemi:
Implements spec `007-darnlink-check` (constitution review: green on substance, renamed `gate`→`check`).

## What
`darnlink check [PATH]` — a **report-only** subcommand that runs **both** axes in one invocation and exits with a distinguishable code:

| exit | meaning |
|---|---|
| 0 | clean |
| 2 | integrity failure (broken/unresolvable robust links or invalid frontmatter) |
| 3 | strict-only failure (anchorable plain links un-anchored) |
| 1 | usage (not a directory) |

Integrity takes precedence when both fail. It **never writes**. This closes the *"strict is not a superset of repair"* trap: a gate that ran only `--robustify` was blind to broken robust links. **darnlink checks + reports; the consumer (CI/hook acting on the exit code) is what gates.**

## How
- `cli.py`: `check` routed as the first token — **backward-compatible**, the existing flag-based CLI is untouched. Report-only (no `--write`). Honours `--exclude` / `--ignore-block` / in-file markers; `--json` separates the two axes.
- `tests/test_check.py`: 8 acceptance tests (broken→2, unresolvable→2, un-anchored→3, both→2 precedence, clean→0, **never-writes** checksum assert, `--json` shape, not-a-dir→1).
- README + CHANGELOG + docstring.

## ⚠️ Deviation from spec FR-001 (flagged for @pma / reviewer)
Spec originally said "build the UUID index once". In the code, each axis scans independently as the existing `darnlink .` / `--robustify` commands do — sharing one index build is a **later optimization**, not required for the value (one command can't-forget-a-half + distinguishable exit code). FR-001 + acceptance wording updated to "single invocation". Please confirm you're OK with this.

Minor known item: argparse-level bad-flag errors exit `2` (argparse default), which overlaps the "integrity" code; the usage cases we control (not-a-dir) return `1`. Can tighten if desired.

## Test plan
- [x] `uv run pytest -q` → 66 passed (8 new)
- [x] dogfood: `darnlink .`, `darnlink . --robustify`, `darnlink check .` → all clean
- [x] `darnlink check . --json` → valid, both axes separated

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #7, opened 2026-07-17
release: v0.3.0

txemi:
Ships `darnlink check` (#6): the report-only gate subcommand (runs both axes, exit 0/2/3/1). Downstream consumers need a tag that includes `check` to pin it (v0.2.0 predates the merge) — this is that tag. Version bump + CHANGELOG only; the feature is already on main and CI-green.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ All notable changes to darnlink are documented here. The format is based on
 
 ## [Unreleased]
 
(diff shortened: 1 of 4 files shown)


txemi:
@Copilot good catch — the README's `darnlink-strict` example still said 'until it's tagged, pin to a main SHA', but v0.2.0 already ships it. Fixed to `rev: v0.2.0   # darnlink-strict ships here`.

PR #10, opened 2026-07-17
feat(recipes): ship darnlink-gate as a public reference recipe

txemi:
Promotes the generic `darnlink-gate` wrapper (bash + `.ps1`) from the **private** toolbelt into `recipes/` in the **public** darnlink repo, so any consumer CI can fetch it **without a token** (fresh CI runners can't reach a private repo; immich is public and can't use a cred).

## Why public / why here
- The recipe is the recommended way to wire darnlink into a repo's gate. Consumers already use the public darnlink; shipping the recipe alongside it makes darnlink **self-contained/self-explanatory**.
- **Constitution:** this is a **reference recipe, NOT part of the CLI/package** — the tool stays "links & UUIDs only"; the script only *orchestrates* darnlink. (Reopens the earlier "orchestration→toolbelt" call; txemi's decision as the tool owner. @pma flagging for your constitution view — is `recipes/` an acceptable reference artifact, core untouched?)

## What
- `recipes/darnlink-gate` (+ `.ps1`) — self-contained (no toolbelt deps), default ref `@v0.4.0`.
- `recipes/README.md` — adoption + the credless CI fetch (`curl .../darnlink/v0.4.0/recipes/darnlink-gate`).
- README pointer + CHANGELOG.

## Test plan
- [x] dogfood `darnlink .` + `--robustify` clean with `recipes/` present
- [x] recipe self-contained (no `source` of toolbelt; only uvx/git/python3)

Follow-ups (separate PRs): toolbelt keeps a **pointer** here + drops its copy; consumers migrate to fetch from here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,14 @@ All notable changes to darnlink are documented here. The format is based on
 
 ## [Unreleased]
 
+### Added
+- **`recipes/darnlink-gate`** — a ready-made, config-driven gate wrapper (bash + `.ps1`), shipped as a
+  **reference recipe** (not part of the CLI/package — the tool stays "links & UUIDs only"). It runs
+  **both** checks, scopes to staged files in pre-commit vs the whole repo in CI, pins the darnlink ref,
+  and fails open on network — so a repo wires darnlink into its gate with a tiny `darnlink-gate.json` +
+  a 3-line hook instead of a bespoke wrapper that drifts. It lives in the **public** repo so any CI can
+  fetch it **without a token**. See `recipes/README.md`.
+
 ## [0.4.0] — 2026-07-17
 
 ### Added
--- a/README.md
+++ b/README.md
@@ -139,6 +139,10 @@ runs it for you.
 darnlink **exits non-zero** when a robust link is broken, so any gate that runs it will block the
 breakage before it lands. Pick the one that fits your workflow — near copy-paste:
 
+> **Want a ready-made wrapper instead of wiring it yourself?** [`recipes/darnlink-gate`](recipes/README.md)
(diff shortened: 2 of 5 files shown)


txemi:
## Verdict on Copilot's review — all findings addressed ✅

Thanks — the review caught a **real** contract bug, not just docs. Fixed in `7250c9b`:

**`mode=repair` now truly means 'integrity only'** (was the core finding, raised on both scopes):
- **repo scope** used to run `darnlink .` (exit 0/1), breaking the advertised `0/2/3` contract → now **always runs `darnlink check`** and maps a strict-only failure (`3`) → `0` when `mode=repair`.
- **staged scope** used to enforce strict findings (exit 3) even under `mode=repair` → now **drops strict findings before gating** when `mode=repair`.
- Same fix mirrored in `darnlink-gate.ps1`.

**Other findings:**
- *Header comment 'bootstraps uv/uvx'* → corrected: it never bootstraps, it **fails open** if uvx is missing. Also fixed the `mode=repair → darnlink .` comment (it's `check` now).
- *`realpath` for the staged filter* → applied (both sides), since darnlink emits resolved absolute paths.
- *`v0.4.0` fetch URL would 404* → already fixed to `v0.5.0` in `f5d2a3c`; **v0.5.0 is the release this PR ships `recipes/` in**, so the URL resolves once merged+tagged (the `darnlink-gate.json` *runtime* ref is also v0.5.0 ⊇ v0.4.0).

**Verified** on a fixture with a strict-only finding (anchorable plain link, no integrity issue): repo/staged × `check` → exit **3**; repo/staged × `repair` → exit **0**. Dogfood + 75 tests green.

PR #14, opened 2026-07-22
feat(recipes): mode=max (fail-closed links) + complete wall examples + doc fixes

txemi:
## What

Adds the **top rung** to the gate (`mode=max`, fail-closed on links) and closes the documentation/example gaps that work uncovered.

## `mode=max` — the third rung of the ratchet (repair ⊂ check ⊂ max)

`darnlink check` stops at **strict** (integrity + robustify against targets that *have* a uuid). It has **no create-frontmatter axis**, so it cannot express *"a link to a file with no uuid fails the gate"* — the fail-closed maximum described in `docs/elevating-your-link-gate.md`.

`mode=max` runs `darnlink . --robustify --create-frontmatter` in **dry-run** (never `--write`) and gates on its exit code. It is a **superset** of `check`, so raising the mode can only tighten, never loosen.

**Whole-repo only**: under staged scope it deliberately stays at strict (the staged pre-commit is meant to be fast; the maximum is enforced at the whole-repo pre-push/CI wall). Ported to `.ps1` as well.

**Verified by running it:**
- One consuming repo (gap 0) → `mode=max` gives **exit 0**.
- Another (+3) → `check` passes (exit 0), `max` **fails** (exit 1, 3 findings).

## Docs fixes (`recipes/README.md` was stale and subtly unsafe)

- It documented neither `fail_closed` (present since v0.6.0) nor exit code 4.
- **The CI example ran `darnlink-gate` WITHOUT `DARNLINK_GATE_FAIL_CLOSED=1`** → copying it gave you a wall that fails open (green build, zero files validated). Fixed.
- The wiring had no **pre-push** layer, although §7 of the playbook defines it as a wall in its own right. Now "4 pieces": pre-commit (staged) · pre-push (whole) · CI (whole, fail-closed).

## `recipes/examples/` (new)

**Complete, copy-paste** artifacts rather than fragments to assemble (assembling the CI one wrong is exactly how you end up with a fail-open wall): `pre-commit`, `pre-push`, a whole Actions workflow (`permissions: contents: read`, `curl -fsSL`, fail-closed) and a Jenkinsfile stage (self-hosted, fail-closed). All pinned to `@v0.7.0` in step with each other.

### DIFF
--- a/README.md
+++ b/README.md
@@ -204,7 +204,8 @@ breakage before it lands. Pick the one that fits your workflow — near copy-pas
 
 > **Want a ready-made wrapper instead of wiring it yourself?** [`recipes/darnlink-gate`](recipes/README.md)
 > does all of the below (both checks, staged-in-pre-commit vs whole-repo-in-CI, pinned ref, fail-open)
-> from a tiny `darnlink-gate.json`. It's a reference recipe, fetchable in CI without a token.
+> from a tiny `darnlink-gate.json`. It's a reference recipe, fetchable in CI without a token — with
+> complete copy-paste hook & CI files in [`recipes/examples/`](recipes/examples/).
 
 **1. pre-commit** (recommended — darnlink ships a hook):
 
--- a/docs/elevating-your-link-gate.md
+++ b/docs/elevating-your-link-gate.md
@@ -191,6 +191,11 @@ When the gap reads **0**, switch your gate command to the maximum:
 
 Re-verify 0, and you're fail-closed: from now on, a link to any file without a `uuid` fails.
 
+> **Using the [`darnlink-gate`](../recipes/README.md) recipe?** This flip is **one line** — set
+> `"mode": "max"` in `darnlink-gate.json`; the hooks and CI don't change. `mode=max` runs exactly the
+> command above (dry-run) at the whole-repo wall, and stays at strict in the staged pre-commit by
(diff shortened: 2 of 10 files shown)


txemi:
## Verdict — Copilot findings addressed + CI note

**Copilot review (all 3 valid, fixed in the latest commit):**
1. ✅ Spanish comments in `recipes/darnlink-gate` → translated to English (fail-closed block, wiring line, EXIT line).
2. ✅ `chmod +x` without a target in the pre-commit/pre-push examples → now names the hook path.
3. ✅ `.ps1` header claiming "always runs `darnlink check`" → already updated to describe `mode=max` in `096c30b`.

**CI — the red `smoke (ubuntu-latest)` is a pre-existing flake, not this change:** the smoke job pins `uvx --from git+…@${{ github.sha }}`, and on a `pull_request` event `github.sha` is the *ephemeral merge commit*, which isn't fetchable → fail. The **push-event run of the same commit `096c30b` passed smoke and every other check**. (Worth fixing the workflow to use `head.sha` on PR events, but that's out of scope here.)

**Two independent adversarial review passes** (subsession): the first caught a critical — `mode=max` was dropping the integrity axis (a fail-open at the wall dressed as the strictest setting); the second confirmed the fix makes `max` a true superset of `check`, verified by executing the recipe on synthetic repos.

PR #22, opened 2026-07-23
feat(recipe): opt-in web + create_readme keys in darnlink-gate.json

txemi:
## What

Adds two opt-in keys to `darnlink-gate.json` (both only under `mode=max`), so a fleet can turn web / create-readme on **by config** instead of wiring each repo by hand:

- **`"web": true`** → a 3rd `web-check --online` pass (robustness of cross-repo links by GitHub URL, anchored with `<!-- web-uuid: X -->`). Public = tokenless; private = `$GITHUB_TOKEN` if present, otherwise `web_unverifiable` (a warning, does not break). **Fail-closed** on a broken public link.
- **`"create_readme": true`** → the `max` pass additionally runs `--create-readme` (directory-link targets need a README with a uuid).

Wired in **both** the bash and the PowerShell recipe. Overrides: `DARNLINK_GATE_WEB` / `DARNLINK_GATE_CREATE_README`.

## Integration bug caught and fixed

`web-check` exits with **code 4** on a broken link (`web_mismatch`), which collided with the recipe's `rc>3 = "darnlink unreachable"` heuristic → **bail fail-open** (green with a broken link). Fix: the web pass **exits with its own code directly**, without going through the bail.

## Test (clean fixture)

| Case | Expected | Actual |
|---|---|---|
| correct web-uuid | 0 | ✅ 0 |
| broken web-uuid | 4 (fail-closed) | ✅ 4 |
| `web:false` (broken link) | 0 (ignored) | ✅ 0 |
| `create_readme:true` with no hubs | 0 | ✅ 0 |

## Why

The consuming fleet's goal: **every repo with the web option on and maximum robustness**, kept in sync by one shared recipe. This is the piece of darnlink that enables it. After merge → release (v0.11.0) → consumers pin it and add `"web": true` to each json.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,19 @@ All notable changes to darnlink are documented here. The format is based on
 
 ## [Unreleased]
 
+### Added
+- **Recipe `darnlink-gate` gains two opt-in `darnlink-gate.json` keys** (both `mode=max` only), so a
+  fleet can turn them on by config instead of hand-wiring each repo:
+  - **`"web": true`** — adds a `web-check --online` pass: cross-repo Markdown links to other GitHub
+    repos must resolve to the destination file's `uuid`. Public targets are tokenless; private ones
+    send `$GITHUB_TOKEN` if set, else report `web_unverifiable` (a warning, never a failure). Fail-closed
+    on a broken public web link. The recipe exits web-check's code directly (its `4` is a real broken
+    link, not the core's `rc>3` "unreachable" — which would otherwise swallow it and go green).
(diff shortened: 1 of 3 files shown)


txemi:
Copilot's finding is valid and fixed in the follow-up commit: the web pass exits web-check's code directly, so the comment now documents the full contract — **0** (clean, incl `web_unverifiable`/offline) · **3** (a plain web link not yet anchored in dry-run — a real gate failure, which is the intent: detect un-robustified web links) · **4** (broken/mismatched public link). All non-zero are fail-closed. Behaviour was already correct (`exit $rc` directly); only the comment was incomplete. CI green on 3.10–3.13. Merging.

PR #25, opened 2026-07-26
feat(recipe): web pass forwards repo excludes to web-check

txemi:
## What
The `darnlink-gate` recipe's 3rd (`web`) pass now forwards to `web-check` the **same `--exclude`** list as the repo (previously only `--ignore-block`).

## Why
`web-check` gained `--exclude` in **0.12.0** (PR #23). Without wiring it through, turning `"web": true` on in a repo that vendors `clones/` or `mirrors/` made `web-check` **fetch (and, with `--write`, anchor) web links INSIDE somebody else's checkout**. Now it skips them just like the core does.

## Test (fixture)
| Case | Expected | Actual |
|---|---|---|
| new recipe (passes `--exclude clones`), broken web link inside `clones/` | 0 (skipped) | ✅ 0 |
| `web-check` WITHOUT `--exclude` on the same tree | 4 (sees it) | ✅ 4 |
| `web-check --exclude clones` | 0 | ✅ 0 |

Bash + PowerShell. After merge → release v0.13.0 → consumers pin it and turn `web` on without scanning vendored clones.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@ All notable changes to darnlink are documented here. The format is based on
 
 ## [Unreleased]
 
+### Changed
+- **Recipe `darnlink-gate`: the `web` pass now passes the repo's `excludes` to `web-check`** (not only
+  `ignore_blocks`). `web-check` gained `--exclude` in 0.12.0; without wiring it, turning `web` on in a repo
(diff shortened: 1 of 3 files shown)


txemi:
Fixed: the Spanish comment in the .ps1 is now English (English-only policy). The bash recipe comment was already English. CI green on all versions.

PR #26, opened 2026-07-26
feat(recipe): web pass reads GITHUB_TOKEN from a PAT file when unset

txemi:
## What
The recipe's `web` pass reads `$GITHUB_TOKEN` from a read-only PAT file (default `~/.config/github_token_ro`, override `DARNLINK_GATE_TOKEN_FILE`) when it is not already in the environment.

## Why
A git hook's environment usually has no `GITHUB_TOKEN`, so turning `web` on in a repo with **private** cross-repo targets would report them as broken (404). With the file, the gate reads it by itself. File absent → private targets stay `web_unverifiable` (does not break), exactly as before.

## Test
Recipe with `GITHUB_TOKEN` unset → reads the file → web-check reads a private target → anchor pending (exit 3). Without the file → not-found. Bash + PowerShell.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,13 @@ All notable changes to darnlink are documented here. The format is based on
 
 ## [Unreleased]
 
+### Added
+- **Recipe `darnlink-gate`: the `web` pass reads `$GITHUB_TOKEN` from a read-only PAT file** when it is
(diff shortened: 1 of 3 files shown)


txemi:
Both fixed: bash now tests `-f` (regular file) so a directory can't slip into the `<` redirection; the .ps1 uses `-PathType Leaf` + try/catch so a read error can't terminate the gate under ErrorActionPreference=Stop (missing/unreadable token just leaves private destinations web_unverifiable). Behaviour unchanged for the happy path.

PR #30, opened 2026-07-30
recipe: create_readme works with mode=check + create_readme_excludes (per-axis) — for repos with a mirror tree

txemi:
## Why

Repos that keep a big **`mirrors/`** tree (a faithful local copy of an external system — an issue
tracker, a chat, a document store) need a gate shape the recipe couldn't express before:

> enforce **robustify** + **create_readme**, but **exclude `mirrors/` from README-creation** (we don't
> invent a README for someone else's export) **WITHOUT** excluding it from integrity/robustify — inbound
> links *into* the mirror must still validate.

Today the recipe either (a) levels down (`mode=check` with no create_readme axis) or (b) breaks
(`mode=max` folds `--create-readme` over the whole tree, so every mirror dir-link floods the
**robustify** axis — on the reference repo that is **268** robustify offenders + 69 create_readme).

## The 2 changes (bash only — the darnlink CLI is untouched)

1. **`create_readme` now works under `mode=check` / `repair`, not only `max`.** It runs as its **own**
   dry-run pass filtered to the `create_readme` findings (reusing the staged scope's JSON-by-kind
   filter), so it adds the folder axis on top of check without inflating the robustify axis.
2. **New `create_readme_excludes` key** (default `[]`): directory globs applied **only** to the
   create-readme pass, layered on top of `excludes`. Suppresses README-creation for those paths
   **without** dropping them from the integrity/robustify axes.

## Backward compatibility

Fully backward compatible. `mode=max` with **no** `create_readme_excludes` keeps the old **folded**
behavior byte-for-byte (create-readme rides inside the max robustify pass); the key absent = empty list.
Existing consumers are unaffected. Covered by two backward-compat tests.

## Tests

New `tests/test_recipe_gate.py` (offline: a `uvx` shim execs the locally installed `darnlink`):

- create_readme fails a README-less dir-link **under `mode=check`** (change 1)
- `create_readme_excludes` suppresses that axis for the excluded dir (change 2)
- excludes are **create-readme-only**: robustify still fires on a plain link *inside* the excluded dir
- backward compat: `mode=max` without the new key unchanged (green when off / red when folded)

`159 passed` (6 new + 153 existing).

## Verification against a real mirror repo (`project_map`)

Config used:

```json
{ "ref": "git+https://github.com/txemi/darnlink@v0.18.0", "mode": "check", "create_readme": true,
  "create_readme_excludes": ["mirrors"], "excludes": [".pytest_cache","clones","output"] }
```

Ran the modified recipe over the `project_map` tree. Result — **NOT 268, no web crash**:

| axis | count |
|---|---|
| integrity (unresolvable) | 7 |
| strict / robustify | 48 |
| create_readme (mirrors excluded) | **3** |
| **total offenders** | **58** |

The `create_readme` axis is **3** with `create_readme_excludes:["mirrors"]` vs **69** without it — and
the robustify axis stays at **48**, never the 268 flood, because create-readme runs as a separate
kind-filtered excluded pass. Exit 2 (integrity), gate correctly red, no `web-check` involvement.

**Honest note on the number:** the old ad-hoc wrapper this ports (`darnlink_robustness_check.py`)
reported **61** = 58 robustify(`--create-frontmatter`) + 3 create_readme, and it never runs the repair
pass so it silently ignores 7 real dangling links. This recipe under `mode=check` reports **58** =
48 robustify (no `--create-frontmatter`, per the config's `mode=check`) + **7 real integrity findings
the old wrapper hid** + 3 create_readme. The −3 delta is entirely that trade (−10 for not forcing
create_frontmatter, +7 for actually enforcing integrity) — i.e. the recipe is stricter/more correct, and
the headline win (create_readme 3-not-69, robustify 48-not-268, no web crash) holds exactly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,19 @@ All notable changes to darnlink are documented here. The format is based on
 
 ## [Unreleased]
 
+### Changed
+- **`recipes/darnlink-gate`: the create-readme axis now works under `mode=check`/`repair` (not only
+  `mode=max`), and gained a per-axis `create_readme_excludes` key — for repos with a big `mirrors/`
+  tree.** With `create_readme: true`, a plain link to a folder with no README now fails the gate under
+  any mode: it runs as its own dry-run pass filtered to the `create_readme` findings (reusing the
+  staged scope's JSON-by-kind filter). The new `create_readme_excludes` (default `[]`) layers extra
+  directory globs **onto the create-readme pass only**, so a repo can skip README-creation under an
+  external mirror (we don't invent a README for someone else's export) **without** dropping the mirror
+  from the integrity/robustify axes — inbound links into the mirror still validate. Fully backward
+  compatible: `mode=max` with no `create_readme_excludes` keeps the old folded behavior byte-for-byte;
+  the key absent = empty. Recipe-only change — the darnlink CLI is untouched. Tests in
+  `tests/test_recipe_gate.py`.
+
 ## [0.18.0] — 2026-07-29
 
 ### Added
--- a/recipes/README.md
+++ b/recipes/README.md
@@ -58,6 +58,30 @@ whole-repo where it's the wall.
 }
 ```
 
+**Repos with a big `mirrors/` tree** (a faithful local copy of an external system) usually want a
+fourth shape: enforce robustify + the create-readme axis, but skip README-creation *under the mirror*
+(you don't invent a README for someone else's export) — while still validating links that point INTO
+the mirror. Two keys make that expressible:
+
+```json
+{
+  "ref": "git+https://github.com/txemi/darnlink@v0.18.0",
+  "mode": "check",
+  "create_readme": true,
+  "create_readme_excludes": ["mirrors"],
+  "excludes": [".pytest_cache", "clones", "output"]
+}
+```
+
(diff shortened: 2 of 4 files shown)


txemi:
### Review verdict

**Copilot review (1 finding) — VALID, fixed.**

- *"If the create-readme pass can't run (python3 missing), `bail` exits 0 in fail-open mode and can mask an already non-zero `rc` from the core check/max pass."* — Correct and important: an **optional** axis that cannot run must never turn an **already-red** gate green. Fixed in `8b8b905`: when the core already failed (`rc != 0`) we exit with that `rc`; only when the core is clean do we defer to the could-not-gate policy (`bail`: fail-open skip / fail-closed 4). Added a regression test that simulates python3-missing via a minimal PATH — it fails without the guard, passes with it.

**CI:** ubuntu + windows green (the bash-recipe tests self-skip on Windows, which ships the separate `.ps1`). `160 passed` locally.

**My own pass:** the two changes are backward-compatible (mode=max without `create_readme_excludes` is byte-for-byte unchanged; key absent = empty). Verified against a real `mirrors/`-heavy repo: `create_readme` axis 3 (excl `mirrors`) vs 69 without it, robustify axis stays 48 (never the 268 flood), no `web-check` involvement.

Ready to merge once CI + Copilot are green.

PR #32, opened 2026-08-07
fix(web-check): summarise web_unverifiable instead of printing one line each

txemi:
## What

`web-check`'s text report printed one line per `web_unverifiable` finding. It now lists
the first 20 and then `... and N more (informational; re-run with --json for the full list)`.

## Why

`web_unverifiable` is informational — it never fails the exit (only `web_mismatch` /
`web_not_found` do). But on a documentation repo whose Markdown holds a few thousand
ordinary external links (docs sites, videos, intranet URLs: anything that is not a GitHub
blob/raw URL), the old report emitted thousands of lines. Two consequences:

1. The actionable findings were buried in the noise.
2. Whoever read the output got flooded. In a `pre-push` git hook — whose stdio is a
   non-blocking pipe — the run died with
   `BlockingIOError: write could not complete without blocking`. So a phase that had found
   nothing wrong (`exit 0` when run standalone) blocked **every push in the repo** until
   `--no-verify` was used, which of course skips the rest of the wall too.

Observed on a repo with ~2200 such links: 2229 lines of output → 24, same `exit 0`.

## Not silenced (Constitution II)

The full count stays in the summary line, and `--json` still carries every finding.
`web_mismatch` / `web_not_found` / `web_anchor` are still listed in full.

## Tests

`tests/test_weblinks.py`: one test asserts the preview + `... and N more` + that the real
total stays visible; another asserts `--json` still returns all N findings.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ All notable changes to darnlink are documented here. The format is based on
 
 ## [Unreleased]
 
+## [0.19.1] — 2026-08-07
+
 ### Fixed
 - **`web-check`: the text report no longer prints one line per `web_unverifiable` finding.** It lists
   the first `UNVERIFIABLE_PREVIEW` (20) and then a `... and N more` line. `web_unverifiable` is
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,21 @@ All notable changes to darnlink are documented here. The format is based on
 
 ## [Unreleased]
 
+### Fixed
+- **`web-check`: the text report no longer prints one line per `web_unverifiable` finding.** It lists
+  the first `UNVERIFIABLE_PREVIEW` (20) and then a `... and N more` line. `web_unverifiable` is
+  informational — it never fails the exit — so on a documentation repo whose Markdown holds a few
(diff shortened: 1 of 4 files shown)


txemi:
## Review verdict

**CI:** 20/20 green on `9802981` — `test` on ubuntu+windows × Python 3.10/3.11/3.12/3.13, plus both `smoke` jobs.

**Copilot review:** 1 finding, **accepted and fixed** in `9802981`.

> The fixtures describe "ordinary external URLs" but carried a `<!-- web-uuid: ... -->` marker, making them anchored links and blurring what the test exercises.

Verified before accepting: in `check_web_links_online`, a link whose `parse_github_url` returns `None` goes straight to `_classify(link, None, 0, None, ...)` — the anchored/plain distinction never enters that path, so the marker changed the reader's understanding and nothing else. Dropping it keeps the fixtures honest about the scenario. Suite still green (162 tests).

No findings dismissed.

**Manual check beyond CI** — ran the patched `web-check` against a documentation repo with ~2200 non-GitHub links: **2229 lines of output → 24**, same `exit 0 (clean)`, actionable findings no longer buried.

Ready to merge from my side.

PR #33, opened 2026-08-08
docs(privacy): use fictitious org/repo names in specs, tests and CHANGELOG

txemi:
Documentation and test fixtures referred to real private repositories and a real third-party organisation by name, plus one internal file path. Public repo, public history — none of it needs to be there, and the examples read exactly the same with placeholders.

## What changed

| File | Was | Now |
|---|---|---|
| `specs/013-web-robustness/spec.md` | two real private repo names + a real internal path | `example-org/handbook`, `ledger`, `docs/living/service-topology.md` |
| `tests/test_weblinks.py` | same names as URL fixtures, plus a real org in the unverifiable-repo test | `example-org/handbook`, `acme-tech` |
| `specs/006`, `specs/010` | "evidence this is real" paragraphs naming the consumer repo | reworded to "a large consumer repo"; the scale figures stay, since those carry the argument |
| `CHANGELOG.md` (0.17.0) | named the consumer repo and the third-party org | reworded generically |

The tests never touch the network (every one injects a fake fetcher), so the URLs are pure fixtures — swapping them changes nothing but the strings. **162 tests pass.**

## Note

This fixes the tree going forward. The names remain in the git history and in the sdists already on PyPI (which ship `specs/` and `tests/`); handling that is a separate decision.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -60,7 +60,7 @@ All notable changes to darnlink are documented here. The format is based on
   otherwise it returns the sentinel `-2` → `web_unverifiable`. The repo probe is cached per
   `(owner, repo, token)` (a repo linked N times is checked once) and falls back to the plain
   404-is-broken behaviour on a network blip. This lets `web: true` run on repos that link to
-  client/third-party orgs (project_map's ~1383 `mapfre-tech` links, etc.) without a wall of false breaks.
+  client/third-party orgs (one consumer repo has ~1400 links into an org it cannot read) without a wall of false breaks.
   Pairs with the token-gated 404 (0.16.0) and the transient-retry (0.15.0).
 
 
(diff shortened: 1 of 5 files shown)


txemi:
Both review comments applied in `a1995cb`:

- **spec 013 line 22** — translated to English (`See the [service topology]`). Valid catch: the line was Spanish and this repo's specs are English; it survived because it was never touched until this scrub.
- **spec 010 line 32** — `8.000`/`15.000` → `8,000`/`15,000`. Same origin: dot separators from the original Spanish-language notes.

Neither was introduced by this PR, but both are on lines it rewrites, so they belong here.

txemi:
**Verdict: merge.**

- CI **20/20 SUCCESS** (Linux + Windows, py3.10–3.13, both smoke jobs).
- Copilot review: 2 comments, **both valid and both applied** (English prose in spec 013, comma separators in spec 010). No findings dismissed.
- No behaviour change — the tests inject a fake fetcher, so the URLs are inert fixtures. 162 tests pass locally and in CI.

Self-approve is blocked on this account, so this comment is the record.

PR #34, opened 2026-08-08
ci: add a privacy gate that fails on private references

txemi:
Follow-up to #33. That PR cleaned the tree; this one stops it happening again.

## Why a gate and not another review

The references removed in #33 did not get in through carelessness with secrets — no credential ever leaked. They arrived as **examples**: a spec documents its motivating case, a PR justifies a change with measured evidence, and the case being described was not public. A one-off manual review at launch cannot catch that, because the next feature re-introduces it. This is the kind of property that has to be checked mechanically on every change.

## Surfaces

| Surface | Why it is checked |
|---|---|
| Tracked files | fixable with a commit, but permanent in history |
| **PR title and body** | **the important one** — public HTML, indexed, and *unretractable*: editing keeps the old revision readable via the API, and a body cannot be deleted. Before merge is the only chance |
| Commit messages | cannot change after push without a rewrite |

## Two deliberate choices

- **Fail-closed on a missing secret.** No patterns → the job fails. A gate that reports green while checking nothing is worse than no gate.
- **Never print the match.** CI logs on a public repo are public; printing the offending text would leak it in the place meant to prevent that. Output is `file:line` only.

The pattern list lives in the `PRIVACY_DENYLIST` secret (one extended-regex per line) for the obvious reason: it cannot live in the repo it protects.

## Verified in both directions

- clean tree → passes
- injected reference → caught, reported as `file:line`, no content in the log
- a PR body carrying a private name → caught
- the anonymised prose from #33 → no false positive

### DIFF
--- a/.github/workflows/privacy-gate.yml
+++ b/.github/workflows/privacy-gate.yml
@@ -58,6 +58,17 @@ jobs:
             exit 1
           fi
           printf '%s\n' "$DENYLIST" | grep -v '^[[:space:]]*$' > /tmp/denylist.txt
+          # Re-check AFTER stripping blank lines, not only before. A secret holding nothing but
+          # whitespace passes the -z test above and then leaves zero patterns, and GNU grep -f with
+          # an empty pattern file matches NOTHING and exits 1 -> every surface would report "clean"
+          # while checking nothing, which is exactly the failure this job exists to make impossible.
+          # (Worth knowing: ugrep does the opposite on the same input, so a local test can disagree
+          # with the runner. Hence an explicit check instead of relying on either behaviour.)
+          if [ ! -s /tmp/denylist.txt ]; then
+            echo "::error::PRIVACY_DENYLIST contains no usable patterns (blank/whitespace only)."
+            echo "Failing closed rather than scanning with an empty pattern set."
+            exit 1
+          fi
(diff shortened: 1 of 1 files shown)


txemi:
Applied in `7c89d13`. **Valid finding, and a hole in the exact property this job advertises** — thank you.

The emptiness check ran *before* blank lines were stripped, so a whitespace-only secret passed it and left zero patterns. Confirmed: GNU grep `-f` with an empty pattern file matches nothing and exits 1, so all three surfaces would have printed `clean` while scanning against nothing.

One detail worth recording, because it nearly made me dismiss this: **ugrep returns the opposite verdict on the same input** (an empty pattern file matches everything, exit 0). My first local reproduction used ugrep via a shell function and appeared to show the gate failing closed already. It was the runner's GNU grep that settled it. That is now the argument for an explicit `[ -s ]` test rather than depending on either implementation, and it is written into the comment so the next person doesn't repeat the mistake.

txemi:
**Verdict: merge.**

- CI **21/21 SUCCESS**, including the gate scanning its own PR (tracked files, title/body and commit messages all clean).
- Copilot: 1 finding, valid, applied in `7c89d13`. It was a hole in the job's central guarantee, not a nitpick.
- Verified the public CI log contains **zero** denylisted terms — the gate reports `file:line` only, never the matching text.

Self-approve is blocked on this account; this comment is the record.

PR #39, opened 2026-08-09
docs: name the consumer monorepo generically in lang_gate

txemi:
A comment in `tools/lang_gate.py` names a private repository. One line, one word.

## How it got in, which is the part worth reading

**The privacy gate caught it.** The run on the push that introduced the file is red — a `failure`, correctly reported.

It landed anyway, because that push went **straight to `main`**. On `pull_request` a red gate blocks the merge; on `push` there is nothing left to block — the commit is already public and the failed run is just a marker someone has to notice. Nobody did. A closing audit found it, not the red run.

So the gate works and its coverage has a hole: it can only *prevent* on the PR path.

## Worth considering separately

Making `main` a protected branch with the gate as a required check would close it properly, by making the direct push impossible. That is a repository-settings change, not something to slip into this PR.


txemi:
**Verdict: merge.** CI 23/23, no review findings. One-word documentation fix.

PR #42, opened 2026-08-09
recipes: the example config recommended two rungs below what it should

txemi:
The `darnlink-gate.json` example in the recipe header still pinned **v0.7.0** — thirteen releases behind — and recommended `"mode": "check"` with **no `dangling` key at all**, while the axis has existed since v0.20.0 and this very file implements it.

A reference config that lags its own tool teaches the wrong thing to everyone who copies it, which is what a reference config is *for*. Found while checking that what a fleet actually applies matches what its own docs recommend — it did not.

**What changes**

- The example now shows the **target**: every axis at its strictest, and says so, so nobody reads it as a minimum.
- **Adoption path added.** Pasting the target into a repo that has never been gated fails on contact, and a gate that fails on arrival gets deleted rather than climbed. The six rungs are now spelled out in order, each fixing content before raising the bar — which is what the one-way ratchet has always meant but never showed.
- The `dangling` rung carries its own sub-ladder, because it is the axis that surfaces the most debt at once: `warn` measures without failing, `added-lines` fails only on lines the commit *adds* (old debt never blocks anyone), `repo` fails on any. Quoted with a real figure — **2,356 links pointing at files that do not exist** across a nine-repo fleet, none named by any other axis — so a reader expects a number instead of being surprised by one.
- A note to pin `ref` to the **latest release** rather than to whatever the comment says: a hardcoded version in an example is exactly how this drifted for thirteen releases.

Docs/comments only; no behaviour change. `tools/check.sh` green (201 tests, link gate clean, lang gate clean).

### DIFF
--- a/recipes/darnlink-gate
+++ b/recipes/darnlink-gate
@@ -33,15 +33,20 @@
 #     findings: 2 integrity / 3 strict).
 #
 # CONFIG. Reads `darnlink-gate.json` at the repo root (all keys optional):
-#   { "ref": "git+https://github.com/txemi/darnlink@v0.20.1",
+#   { "ref": "git+https://github.com/txemi/darnlink@vX.Y.Z",
 #     "excludes": ["secrets","node_modules",".venv"], "ignore_blocks": ["txmd-autogrid"],
 #     "mode": "max", "scope": "repo", "fail_closed": true,
 #     "web": true, "dangling": "repo", "create_readme": true,
 #     "create_readme_excludes": ["mirrors"] }
 #
-#   ^ That is the TARGET, not a starting point: every axis at its strictest. Pin `ref` to the
-#   LATEST release, not to whatever this comment happens to say -- this example sat at v0.7.0 for
-#   thirteen releases and quietly recommended two rungs below what its own author was running.
+#   ^ That is the TARGET, not a starting point: every axis at its strictest.
+#
+#   `vX.Y.Z` is a PLACEHOLDER ON PURPOSE, and it is the one thing here you must not copy verbatim.
+#   Resolve it when you paste:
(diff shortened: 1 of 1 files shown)


txemi:
## Verdict — both findings accepted and fixed in 3fe1ae9

**1. Hardcoded `ref` in the example (lines 36/42/44) — valid, and uncomfortably so.**

The previous commit was *about* this example being thirteen releases stale, and it "fixed" it by writing another concrete tag, restarting the same clock. The example now reads `@vX.Y.Z` with the command that resolves it (`gh release view -R txemi/darnlink --json tagName -q .tagName`).

It is deliberately no longer copy-pasteable as-is: any pin printed in a comment ages the moment it is committed, so none is printed.

**2. `added-lines` depends on `scope` (line 57) — valid, and it reached further than the doc.**

Verified in this script (lines 336-356): with `scope=repo` there is no staged diff to judge, so `added-lines` degrades to `warn`; only `dangling=repo` fails.

The consequence is not just wording. In the fleet that uses this recipe, three repos sit at `added-lines` (395, 111 and 102 dangling links) and were being described as *"protected against new ones"*. They are — **but only via the pre-commit hook**, which additionally fails open when `uv` or the network is missing. On pre-push and CI that axis is a counter, not a wall: one `--no-verify`, or one unprovisioned machine, and a new dead link lands unopposed.

The rung note now says so, and names the only rung with server-side enforcement: `repo`.

**No suppressed comment left unaddressed** — the third one (line 44) is the same issue as #1 and is covered by the placeholder rewrite.

CI green (201 tests, lang-gate, privacy-gate). Self-approve is blocked on this account, so this comment is the record.

PR #40, opened 2026-08-09
gate: the command the error suggests must be safe to paste

txemi:
## What happens

The strict failure told you to run `darnlink . --robustify --write`. That is correct as a
capability and a trap as advice, for two reasons that only show up once you obey it:

1. It rewrites the **whole repo**, not the file it just complained about.
2. It does **not** carry the `--exclude` list the gate itself was configured with, so it applies a
   *different policy* than the one that just failed you.

Measured in a consuming monorepo: pasted verbatim, it anchored **110 links across 26 files** when
**one** was asked for — on a tree shared by several concurrent sessions. The person who pasted it
did the reasonable thing: they trusted the message.

## What this change does

Suggests the piped `--only-from -` form that this project's own CLI help already recommends: scan
the repo (the link's **target** has to be seen for the anchor to resolve) but write only the staged
files.

```
git diff --cached --name-only -- '*.md' | \
  uvx --from <ref> darnlink . --robustify --create-frontmatter --write --only-from -
```

It also spells out `--create-frontmatter`. Without it, darnlink silently skips a target that has no
frontmatter yet: it reports `skipped (no frontmatter)`, writes nothing, and the anchor never
appears — a "did nothing" that reads as "nothing to do". That cost two attempts to diagnose.

And it says, in one line, not to run the bare global form. **A gate that dictates an action owns the
consequences of that action.**

## Verification

End to end, not by reading:

1. Reproduced the strict failure in the consuming monorepo (`DARNLINK_GATE_SCOPE=staged`).
2. Pasted the new suggestion **verbatim**.
3. Result: `WROTE 1 file(s)`, and the gate turned green (`rc=0`).

Before the change the same situation produced 26 modified files.

<sub>Translated from the original Spanish on 2026-08-11 — this repo is English-only on every surface, PR titles and descriptions included.</sub>

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #43, opened 2026-08-09
release: v0.20.2 — the safe-to-paste gate suggestion finally reaches somebody

txemi:
Cuts **v0.20.2**. Four commits have been sitting on `main` since yesterday with no way to reach any machine: consumers pin the tag, not `main`, so **merged is not the same as delivered**.

The one that was overdue is not the cosmetic one:

- **#40** — the staged gate suggested a command that rewrites the **whole repo** and, worse, does **not** carry the `--exclude` list of the config that had just failed you, so it applies a different policy than the one that produced the error. Measured in a consuming monorepo: **110 links anchored across 26 files when one was asked for**, pasted by someone who trusted the message, on a tree shared by several concurrent sessions. That fix has been merged for a day and **nobody has it**.
- **#42** — the reference config becomes the strict target, with an adoption path, and the `ref` is a deliberate `vX.Y.Z` placeholder: any concrete pin written into a comment ages the moment it is committed, which is exactly how this example drifted for thirteen releases.
- The English-only ratchet gate with its per-file baseline.

**No GitHub Release will be published**, only the tag — publishing one triggers `publish.yml` → PyPI, which is irreversible and was not requested. Same as v0.20.0 and v0.20.1.

Version bump + CHANGELOG only; no behaviour change in this PR. `tools/check.sh` green (201 tests, link gate, lang gate).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,38 @@ All notable changes to darnlink are documented here. The format is based on
 
 ## [Unreleased]
 
+## [0.20.2] — 2026-08-09
+
+### Fixed
+- **`darnlink-gate`: the command the failure message tells you to paste no longer rewrites the whole
+  repo under a different policy** (#40). The staged gate used to suggest a bare
+  `darnlink . --robustify --write`: correct as a capability, a trap as advice. It touches every file
+  in the tree, and it does **not** carry the `--exclude` list the gate that just failed you was
+  configured with — so it applies a different policy than the one that produced the error. Measured
+  in a consuming monorepo: **110 links anchored across 26 files when one was asked for**, pasted by
+  someone who trusted the message, on a tree shared by several concurrent sessions. The suggestion is
+  now the piped `--only-from -` form (scan the repo, write only the staged files), carries the run's
+  own `--exclude`/`--ignore-block` shell-quoted, spells out `--create-frontmatter` (without it the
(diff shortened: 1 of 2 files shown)


txemi:
## Verdict — the finding is valid, and it was bigger than this PR (fixed in 39b0221)

**`## [0.20.2]` is a reference-style link with no definition** — correct, and not a problem with my header: **28 version headings, 6 link definitions**. The footer stopped being maintained at `0.5.0`, and `[Unreleased]` was still comparing `v0.5.0...HEAD` — twenty-two releases behind.

It is the same silent-rot shape this repo has been fixing elsewhere for two days: a reference-style link with no definition **does not error, it renders as plain text**. Nothing looks broken; there is simply no link. That is why it lasted.

Regenerated the whole footer, chained as `compare/vPREV...vCUR`.

**Regenerating it surfaced the interesting part: `0.19.2` has a CHANGELOG section and was never tagged.** The next release commit went straight to `0.20.0`, so that fix (#35, the duplicated uuid on robustify) shipped inside **v0.20.0**. Generating the definitions blindly would have asserted a `v0.19.2` tag that does not exist *and* broken the `0.20.0` link, which must compare against `v0.19.1`.

The section stays — the change is real and dated — with a note that no such release exists and a link to the commit instead of a phantom tag. Deleting it would lose the bug write-up; folding it into `0.20.0` would falsify the date.

`tools/check.sh` green.

PR #44, opened 2026-08-09
gates: add the missing pre-push surface and dogfood the dangling axis

txemi:
This repo asks its consumers for four gate surfaces and for `dangling` at its strictest rung. It was running with three surfaces and the axis off.

**`pre-push` — the surface that was missing.** There was pre-commit and there was Actions, and nothing in between. `git commit --no-verify` is one keystroke, and a clone that never activated hooks runs no gate at all; in both cases the first wall the work meets is CI — remote, slower, and occasionally down. `git push` is deliberate and infrequent, so a whole-repo judgment there is affordable, and it is the last place something broken can be stopped before it leaves the machine. It runs `tools/check.sh`, the same gate as pre-commit and CI, so no drift between surfaces.

**`dangling`, fail-closed at zero.** A dangling link is invisible to every other axis: not `unresolvable` (that is a robust link whose uuid died), not `to robustify` (there is nothing to anchor to). `check` already reports it, but only as informational — the *policy* lives in `recipes/darnlink-gate`, and the CLI has no flag for it by design (recipe owns policy, library owns findings). So the self-gate makes the same judgment from the same JSON, in `tools/dangling_gate.py`, sibling to `lang_gate.py`. Zero baseline because the repo measures zero: a ratchet that costs nothing is one you take before the debt arrives. Four repos in the consuming fleet already run this rung.

**And there was exactly one — illustrating the very failure the axis is for.** `specs/011-directory-links/spec.md` explained the feature with *"see the [deployment guide](ops/deploy/)"*, written as a real link to a path that has never existed here. **An illustration must not be a link**: a reader who clicks it gets a 404 from our own documentation. It is now a code span, which is what it always was semantically.

Verified the gate actually bites: clean tree → exit 0; inject a link to a non-existent path → exit 1, naming file and line.

### DIFF
--- a/hooks/pre-push
+++ b/hooks/pre-push
@@ -20,5 +20,11 @@
 # (lang gate, tests, darnlink self-check at mode=max, dangling at zero).
 #
 # Bypass (discouraged, and it skips the wall for EVERY commit in the push): git push --no-verify
+#
+# stdin is redirected from /dev/null ON PURPOSE. git feeds a pre-push hook the list of refs being
+# pushed on stdin, and anything down the chain that reads stdin -- even incidentally, as `gh` does
+# inside a `while read` loop -- would silently swallow those lines and behave as if the input were
+# empty. That failure mode is invisible: no error, just a wrong answer. Nothing here needs the ref
+# list, so the safest thing is to not hand it over.
 set -euo pipefail
-exec bash "$(git rev-parse --show-toplevel)/tools/check.sh"
+exec bash "$(git rev-parse --show-toplevel)/tools/check.sh" < /dev/null
--- a/hooks/pre-push
+++ b/hooks/pre-push
@@ -1,0 +1,24 @@
+#!/usr/bin/env bash
+# Versioned git pre-push hook — the WHOLE-REPO wall, and the surface this repo was missing.
+#
+# Activate ONCE per clone (git will not auto-run versioned hooks, by design — security):
+#   git config core.hooksPath hooks
+# (Or let the toolbelt's global dispatcher do it: it runs hooks/<name> and hooks/<name>.d/*.)
+#
+# Why a pre-push when pre-commit already runs the same gate.
+# `git commit --no-verify` is one keystroke, and a machine that has never activated hooks runs no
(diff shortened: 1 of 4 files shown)


txemi:
## Verdict — all three accepted and fixed in ff22edf, and testing them surfaced a fourth that was mine

**1. `pre-push` passing its stdin through — valid.** git hands the hook the list of refs on stdin; anything downstream that reads stdin swallows it and behaves as though the input were empty, with no error, just a wrong answer. Not theoretical: earlier today, auditing branches, `gh` ate the stdin of a `while read` loop and made me conclude that **no branch had a PR**, which was false end to end. Nothing here needs the ref list, so it is no longer handed over.

**2 & 3. The gate must say *why* it could not judge — valid.** With no JSON, the message was identical for every cause and omitted the exit code, which is the fastest discriminator (127 = no `uv`, 126 = permissions, 1/2 = darnlink refusing). With unparseable JSON, whatever *did* come out was dropped — and that is the only thing identifying the cause, usually a traceback or warning printed ahead of the payload. Both streams are now shown, and the helper says **"stream was empty"** explicitly: an empty stream printed as nothing is indistinguishable from the gate forgetting to print it, and "empty stderr with a non-zero exit" is itself a clue that points at the wrapper rather than at darnlink.

**4. The one that appeared while testing your findings.** Simulating the most likely breakage — a machine without `uv` — the gate produced no diagnostic at all: it died with a **raw traceback and exit 1**. A missing `uv` does not come back as a return code; the exec itself fails, so `subprocess` raises before there is anything to inspect, and my fix for (2) was reading `proc.returncode` on a `proc` that never existed.

Exit 1 is worse than ugly there: it means *"the gate found something"* when the truth was *"the gate could not run"* — the distinction the `2` code exists for. Now `OSError` is caught, with a clear message and the install command. It is also the first scenario an unprovisioned machine hits, i.e. the common case rather than the exotic one.

All three paths verified by **actual exit code**, not by what the screen printed: clean → 0, no `uv` → 2, injected dangling link → 1.

CI green. Self-approve is blocked on this account, so this comment is the record.

PR #45, opened 2026-08-10
gate: the web pass exited the script and skipped every axis after it

txemi:
The `web` pass ended in a bare `exit "$rc"`. That leaves the script **before** the `create-readme` axis further down, so in a repo with `mode=max` + `web: true` + `create_readme_excludes` **that axis never ran**: the config said it was on, the gate answered `exit 0`, and a directory link to a folder with no README sailed straight through.

Measured on two consuming repos before the fix — same tree, same config, opposite verdicts:

```
web: true    ->  exit 0   (the injected link passes)
web off      ->  exit 1   (caught)
```

One of them had spent half a day with the axis nominally "on" and gating nothing.

## Why the `exit` was there, and what is preserved

It was not gratuitous. `web-check`'s exit codes are all in 0..4 and **none of them means "could not run"**, so they must **not** go through the `rc>3 → bail` heuristic below, which would read a genuine `4` (broken public web link) as a network hiccup and go green.

The fix is not to leave early: it marks the verdict as **already final** (`RC_IS_FINAL`) and falls through. `rc` already carries the value, later axes only ever raise it from 0, and the exit happens once at the end with the worst verdict.

## Tests

Two, and the first was validated the way a regression test should be — **it fails without the fix and passes with it** (checked by `git stash`-ing the recipe):

- `test_web_pass_does_not_skip_the_create_readme_axis` — with web on, the axis still runs and fails.
- `test_web_pass_verdict_survives_the_later_axes` — and falling through neither downgrades a web failure nor dirties a clean tree.

They run **offline**: the sandbox has no GitHub links, so `web-check --online` returns 0 without touching the network. What is under test is the fall-through, not the web check.

`tools/check.sh` green (203 tests).

### DIFF
--- a/recipes/darnlink-gate
+++ b/recipes/darnlink-gate
@@ -330,16 +330,28 @@ if [ "$SCOPE" != "staged" ]; then
       # failure is reported web_unverifiable, still exit 0 — it NEVER exits high); 3 = a plain web link
       # not yet anchored (dry-run: it COULD be robustified) — a real gate failure, which is the point
       # (detect un-robustified web links); 4 = a broken/moved/mismatched public web link. So EVERY
-      # non-zero is fail-closed, NOT the core's rc>3 "unreachable" case — exit it DIRECTLY, bypassing the
-      # bail heuristic below (which would swallow a genuine 4 as "network hiccup" and go green).
-      exit "$rc"
+      # non-zero is fail-closed, NOT the core's rc>3 "unreachable" case: it must NOT go through the
+      # `rc>3 -> bail` heuristic below, which would swallow a genuine 4 as "network hiccup" and go green.
+      #
+      # ⚠️ This used to be a bare `exit "$rc"`, and that was a BUG: it left the function before the
+      # create-readme axis further down, so in a repo with `mode=max` + `web: true` +
+      # `create_readme_excludes` the create-readme axis NEVER RAN. The config said the axis was on, the
+      # gate reported a clean exit 0, and a directory link to a folder with no README sailed through.
+      # Measured on two consuming repos before the fix: injecting exactly that case returned 0 with web
+      # on and 1 with web off — same tree, same config, opposite verdicts.
+      #
+      # So: remember the web verdict, mark it as already-final (no bail heuristic), and FALL THROUGH so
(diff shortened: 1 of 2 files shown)


txemi:
## Verdict — the three findings accepted and fixed in `5d1fe89`. The third was a bug in my own fix

**1. `RC_IS_FINAL` had no ceiling — this one matters most.** It skipped the `rc>3 → bail` heuristic *always*, not only for the contract's `4`. The immunity is sound inside `web-check`'s contract (codes 0..4, none meaning "unreachable", so a `4` is a real verdict and not a network hiccup). But a code **above** the contract — `127` no `uvx`, `126` permissions, a runtime blowing up mid-pass — **is not a verdict about the repo**: the tool did not run, and it must go through `bail()`.

Without the ceiling, a machine without `uv` would have received a hard `127` from a gate that promises to **fail open**. So while fixing a silent failure I was introducing a loud one: from *"doesn't gate and says nothing"* to *"blocks whoever can't gate"*. Now `>4` always bails, `RC_IS_FINAL` or not.

Locked with a test, validated the way a regression test should be — **it fails on the previous code and passes on the new one**: a shim that answers normally for every pass and dies with `127` only for `web-check`, which is exactly the narrow window (core passed, then the tool became unreachable). It asserts both policies: fail-open **skips with 0**, fail-closed **exits 4**.

**2 & 3. Spanish in an English-only repo** — an inline comment in the recipe and two test placeholders. Both fixed.

Worth noting *why* they slipped: **`tools/lang_gate.py` only inspects `.py` files.** Bash comments in `recipes/` and string literals in tests are outside its reach, so the English-only rule is enforced over part of the repo and trusted over the rest. Not fixing that here — it is a separate change and deserves its own PR — but it is a real gap in the gate, and this PR is evidence of it.

## Unrelated: the Windows smoke job is flaky by design

It failed on this PR and **passed on a re-run with no change**. The job pins `github.sha`, which on a `pull_request` event is the **synthetic merge commit**; `uvx --from git+…@<sha>` cannot always resolve it once the base has moved. It passed on #42–#44 and lost the race here. Also worth its own PR — it is a green/red that depends on the clock, not on the code.

`tools/check.sh` green (204 tests).

PR #47, opened 2026-08-10
gate: 'dangling_max' — a budget that makes the 'repo' rung adoptable before you reach zero

txemi:
## The gap this closes

The `dangling` axis has four rungs — `off` · `warn` · `added-lines` · `repo` — and a hole between the last two.

`added-lines` is the adoption rung: old debt never blocks, new debt cannot enter. But it **needs a staged diff**, so it only ever bites on `scope=staged` — the pre-commit hook. On a whole-repo surface (`scope=repo`: pre-push, CI) there is no diff to judge and it degrades to `warn`.

The consequence is the part worth stating plainly:

> **A repo carrying old dangling debt has NO server-side wall for this axis, on any surface, until the day it reaches exactly zero.**

And the only local surface that does gate it fails open when `uv` is missing. In a consumer measured while writing this, that meant **256 dangling links** and no server wall — months of "the axis is on" that in practice meant *a habit, not a wall*. A regression could land in `main` and nothing would go red.

## What this adds

`"dangling_max": <int>` (default `0`) turns `repo` into a **budget**: fail only above that count.

```json
{ "dangling": "repo", "dangling_max": 141 }
```

You get the wall **today**, at your current number, and every cleanup lowers it. Same shape as a coverage floor — the number is a receipt of where you are, not permission to stay.

## The three details that make it a ratchet and not an allowance

**1. Coming in under budget is reported, loudly.**

```
darnlink-gate: 118 dangling link(s), under the budget of 141 — lower dangling_max to 118 to keep the ratchet.
```

A budget nobody lowers *is* an allowance, and **nothing else in the system can notice it went stale**. So it has to be said here, in front of whoever just did the cleanup — the one moment someone is both looking and able to act.

**2. A junk value fails closed, not open.**

`"dangling_max": "muchos"` → warns and uses `0`. A typo must never silently **widen** an allowance; that is the one direction a config error must not be able to go.

**3. Absent key = byte-identical verdict.**

No `dangling_max` and `repo` still fails on any finding, and the output does not mention a budget nobody set. Raising the pin changes nothing for anyone who did not opt in — the same rule the whole axis was built on.

The budget applies to `repo` only, since that is the only rung with a wall.

## Tests

6 new, `22/22` green in `tests/test_recipe_gate.py`:

- default (no key) fails on any finding **and stays silent about budgets**
- over budget → fails, naming the budget
- exactly at budget → passes
- under budget → passes **and asks for the budget to be lowered**
- junk budget → fails closed, saying why
- a budget set on the `warn` rung does not turn it into something that fails

One of them caught a bad assertion of mine: the sandbox path carries the test name, so a bare `"budget" not in output` matched the *path*, not the message. Now it matches the phrases.

### DIFF
--- a/recipes/darnlink-gate
+++ b/recipes/darnlink-gate
@@ -147,7 +147,25 @@ case "${CREATE_README,,}" in ""|0|false|no|off) CREATE_README="" ;; *) CREATE_RE
 #                one line of a README that carries old danglers would block the commit, and a gate
 #                people bypass gates nothing;
 #   repo         fails on ANY finding. The wall, for a repo already at zero.
+#
+# `dangling_max` (int, default 0) turns `repo` into a BUDGET: fail only above that count. It exists
+# because between `added-lines` and `repo` there is no rung — and `added-lines` needs a staged diff,
+# so it only ever bites in pre-commit. A repo with old debt therefore has NO server-side dangling
+# wall at all, on any surface, until the day it reaches exactly zero. That is a long time with no
+# wall, and it is precisely when a regression slips in unseen.
+# With a budget you get the wall TODAY at your current number, and every cleanup lowers it. Same
+# shape as a coverage floor: the number is a receipt of where you are, not permission to stay.
+# To keep it a ratchet and not a permanent allowance, the gate SAYS SO when you come in under
+# budget — an unlowered budget is the failure mode, and nothing else would catch it.
+# Only `repo` has a wall, so the budget only applies there. A non-numeric value is treated as 0
+# (strictest) on purpose: a typo must never silently WIDEN an allowance.
 DANGLING="${DARNLINK_GATE_DANGLING:-$(read_cfg dangling off)}"
+DANGLING_MAX="${DARNLINK_GATE_DANGLING_MAX:-$(read_cfg dangling_max 0)}"
+case "$DANGLING_MAX" in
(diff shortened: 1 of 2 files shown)


txemi:
## Verdict: ready to merge

**Copilot's finding was a real bug, not a nit.** The budget verdict lived inside the `there are findings` branch, so the gate went **silent at exactly `dl_n=0`** — the one moment the stale budget is both visible and free to remove. A ratchet whose reminder disappears on success is an allowance with a grace period. It also caught the matching test gap.

Fixed in `d82d106`: the verdict now runs at zero too (only when a budget is set), with its own message for the milestone — *"no dangling links left — drop dangling_max so the wall is a wall again"*.

**Two tests added, and the first one was validated by failing it against the previous recipe** — otherwise it proves nothing:

```
old recipe -> FAILED test_reaching_zero_with_a_budget_still_asks_for_it_to_be_dropped
new recipe -> passed
```

The mirror case is covered too: a clean repo with **no** budget says nothing at all.

**Checks:** 24/24 green, `mergeStateStatus: CLEAN`.

**Process note:** `d82d106` lands after Copilot's review, which only fires when the PR opens. It is the fix for its own finding plus the test it asked for — not re-requesting a review for that.

Self-approve is blocked by GitHub; this comment is the verdict.

PR #48, opened 2026-08-10
release: v0.20.4 — dangling_max, the rung that was missing between added-lines and repo

txemi:
Release of `dangling_max` (#47), which without a tag is unreachable for consumers: they all pin by `vX.Y.Z` in their `darnlink-gate.json`.

## What goes in

`"dangling_max": <int>` (default `0`) turns the `repo` rung of the `dangling` axis into a **budget**: it only fails above that number.

It exists because between `added-lines` and `repo` there **was no rung**. `added-lines` needs a staged diff, so it only bites on `scope=staged`; over a whole repo it degrades to `warn`. The consequence, stated plainly:

> A repo carrying old debt **has no server-side wall for this axis, on any surface**, until the day it reaches exactly zero.

Measured on a consumer: the axis had been **"on" for months** and its only wall was a local hook that itself fails open without `uv`. Raising the setting was no escape either — `repo` demands zero and the repo was at 256.

## Compatibility

**Without the key, the verdict is byte-for-byte identical.** Bumping the pin changes nothing for anyone who did not opt in — the rule the whole axis was built on. A test pins this, including that the output does not mention budgets nobody set.

## Tests

**212 green**, 8 new in `tests/test_recipe_gate.py`. Two of them cover the case that slipped into the first draft and was caught in review: on reaching **zero** the gate went silent, exactly when a stale budget is visible and free to remove. Its regression test was validated by **failing against the previous recipe**.

## After this

The fleet's deploy script (`DARNLINK_GATE_VERSION`) and the `darnlink-gate.json` files across consumers can move to `v0.20.4` when convenient. None of them has to: without `dangling_max` they behave exactly as today.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,40 @@ All notable changes to darnlink are documented here. The format is based on
 
 ## [Unreleased]
 
+## [0.20.4] — 2026-08-10
+
+### Added
+- **`darnlink-gate`: `dangling_max`, a budget that makes the `repo` rung adoptable before you reach
+  zero** (#47). The `dangling` axis had four rungs and a hole between the last two. `added-lines` —
+  the adoption rung — **needs a staged diff**, so it only ever bites on `scope=staged`. On a
+  whole-repo surface (pre-push, CI) there is no diff to judge and it degrades to `warn`.
+
+  The consequence is worth stating plainly: **a repo carrying old dangling debt has no server-side
+  wall for this axis, on any surface, until the day it reaches exactly zero.** Measured on a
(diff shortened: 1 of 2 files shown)


txemi:
## Verdict: ready to merge

**Copilot reviewed both files and generated no comments.** Release bookkeeping only — version bump + changelog entry + compare links.

**Checks:** all green, `mergeStateStatus: CLEAN`. Full suite `212 passed` locally before the branch was cut.

The behaviour this releases (#47) was reviewed on its own PR, where Copilot's one finding — the budget reminder going silent at `dl_n=0` — was a real bug and is fixed here.

Self-approve is blocked by GitHub; this comment is the verdict.

PR #51, opened 2026-08-11
spec 016: fail on web links to repos you own whose destination has no uuid

txemi:
## What

Specification only — no implementation in this PR.

The web axis (feature 013) is forgiving by design: a destination that fetches 200 but carries no `uuid` is reported `web_unverifiable` and the run still exits 0. That is correct when the destination lives in someone else's repository, and wrong when it is **yours** — there it is not an external limitation, it is a missing two-line edit in a repo you control, and nothing ever tells you so.

This spec adds an opt-in `--own OWNER` (repeatable, plus a separate `--own-from-origin`), a new `web_own_no_uuid` finding at **exit 4**, a `web_own_exempt` escape hatch, and a `--own-max` budget so the axis can be adopted before a repo reaches zero.

## Why exit 4 and not 3

Exit 3 promises "re-run with `--write`". darnlink **cannot** fix this one: the edit belongs to the destination repository, and Principle II forbids writing there.

## Two exclusions, and the one that was got wrong twice

- A destination that is **not `.md`** can never carry frontmatter.
- A destination pinned to a **commit SHA** (`^[0-9a-fA-F]{7,40}$`, case-folded) can never be given one retroactively.

The exclusion is **textual** and stops there. An earlier revision widened it to "any ref that is not a moving branch head", to cover tags — wrong, and the spec now records why: a tag `v1.2.3` and a branch `v1.2.3` are textually identical, so honouring it needs a network call, and its failure direction is the bad one (a maintenance branch excluded for *looking* like a tag = a false green, the exact failure this feature exists to remove).

## The budget could not be built the obvious way

`--own-max` lives in **darnlink**, not in a gate config. A recipe re-counts `dangling` findings itself and computes its own exit code; the web pass takes `web-check`'s verbatim, and exit 4 is *shared* with `web_mismatch`/`web_not_found` — so a gate cannot tell a budgeted finding from a genuinely dead link. **The gate wiring is deliberately out of scope**: three review rounds spent their findings on it, it is about how a gate consumes any darnlink exit code, and it is how `dangling_max` was done — axis first, recipe key in its own change.

## Evidence

Measured across a nine-repository fleet, in one stated scope: **255** GitHub `blob`/`raw` links, **228 (89%)** pointing at repos owned by the same account, **193 already anchored**. Of the 35 still plain, **17 point at a `.md` whose frontmatter has no `uuid`** — invisible to every axis today.

Two caveats stated rather than buried: the snapshot is **post-remediation** (11 links were fixed by hand that day so the axis can open with a small budget), and **every ref in the fleet is a branch** — so the immutable-ref exclusion is reasoning, not measurement, and the spec says so instead of claiming both exclusions were measured.

## Constitution

The explicit `--own` form costs **nothing**: a textual predicate over data the run already holds, no extra fetch, nothing stored as an index. Only `--own-from-origin` needs a named sentence in P-IV, because it reads the clone's git configuration.

## Review

Copilot returned an empty review (quota) on every push. **Seven independent adversarial passes** were run instead; every verdict and everything applied is in the PR comments. They caught, among others, a correction that had *broken* the rule it was fixing, an escape hatch that did not escape, and a marker whose behaviour contradicted the spec's own "byte-identical" promise. The seventh found no high-severity issue, which is the agreed bar for merging.

The prerequisite this spec was blocked on — `web-check` crashing on an href it cannot send — **landed in `main` as #55** while this was in review.

### DIFF
--- a/specs/016-own-repo-web-strictness/spec.md
+++ b/specs/016-own-repo-web-strictness/spec.md
@@ -40,68 +40,94 @@ ownership costs **no extra fetch**: it is a textual comparison on a value the ru
 
 ## Evidence this gap is real (measured, not hypothetical)
 
-Measured 2026-08-11 across a nine-repository private fleet, all of which already carry web links:
+Measured 2026-08-11 across a nine-repository private fleet, all of which already carry web links.
+Scope of the count, so it is reproducible: Markdown outside vendored `clones/` and `mirrors/`, with
+fenced code blocks excluded.
 
-- **345** GitHub `blob`/`raw` links in Markdown; **265 (77%)** point at repositories owned by the
-  same account as the linking repo. Cross-repo linking here is overwhelmingly *self*-linking.
-- Of the **46** own links still plain, resolving each destination against a local checkout:
-  **27 point at a `.md` whose frontmatter has no `uuid`** — every one of them invisible to every
-  darnlink axis today. Only 7 would already be caught as `web_anchor`. The rest of the 46: 2 target a
-  non-`.md` file, 2 are pinned to a commit SHA, 1 target no longer exists, and for 9 the destination
-  repository was not checked out locally, so they could not be resolved offline at all.
-- Restricted to the repos that actually run the web axis today (`web: true`), the numbers are
-  **11 links** blocked on **8 destination files** across 3 repositories. All 8 were hand-written
-  documents; none was machine-generated.
+- **255** GitHub `blob`/`raw` links; **228 (89%)** point at repositories owned by the same account as
+  the linking repo. Cross-repo linking here is overwhelmingly *self*-linking.
+- **193 of those 228 are already anchored.** The axis is in daily use, not a proposal.
+- The **35** still plain break down like this — one bucket each, in this order:
+
+| Bucket | Count | Meaning |
+|---|---|---|
+| destination `.md` **without** `uuid` | **17** | invisible to every darnlink axis today — *this feature* |
+| destination has a `uuid` | 5 | already caught as `web_anchor` where the axis is on |
+| destination is not `.md` | 6 | can never carry frontmatter (FR-005) |
+| destination repo not checked out locally | 7 | could not be resolved offline for this count |
+| ref pinned to a commit SHA | 0 | none in the current snapshot (see FR-006) |
+| destination file missing | 0 | would be `web_not_found`, not this |
 
 Two secondary measurements shape the design:
 
-- **In repos where the web axis is already on and fail-closed, the "destination has a uuid" column is
-  zero.** That is structural, not luck: such a link fails the gate as `web_anchor`, so it gets
+- **In repos where the web axis is already on and fail-closed, the "destination has a uuid" bucket is
+  empty.** That is structural, not luck: such a link fails the gate as `web_anchor`, so it gets
   anchored immediately. Everything still plain in those repos is, by construction, in the silent
-  bucket this feature targets. The axis has no overlap with what already exists.
+  bucket this feature targets. The axis has no overlap with what already exists. (The 5 above all sit
+  in repos that have not switched the axis on.)
 - **The practice is growing fast.** In the largest consumer, own web links went 1 → 3 → 3 → 39 →
-  **191** over twelve months (146 of them anchored). A 27-link backlog is not the point; the open
-  door is.
+  **191** over twelve months. A 17-link backlog is not the point; the open door is.
 
-One repository in the fleet mixes **26 own plain links with 30 third-party ones in the same tree** —
-the clearest statement of the problem: to tolerate the 30 you must today also tolerate the 26.
+Two caveats stated rather than buried, because the whole case rests on these numbers:
+
+- **The snapshot is post-remediation.** On the day of measurement, 11 of these links were fixed by
+  hand — `uuid` added to 8 destination files, then anchored — precisely to let the new axis start at
+  zero. Before that day the uuid-less bucket was 27.
+- **Every ref in the fleet is a branch**: `main` 201, `master` 20, `HEAD` 7, and nothing else across
+  all 228. The immutable-ref exclusion below is therefore **reasoning, not measurement** — it has zero
+  occurrences today and is specified so the first one does not become an unfixable failure.
 
 ---
 
-## Decision: an explicit owner list, `auto` as a convenience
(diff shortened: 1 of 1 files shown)


txemi:
## Review verdict (independent pass — Copilot's review came back empty: quota exhausted)

`reviews=1`, `mergeStateStatus: CLEAN` and every check green, but the single review body reads *"unable to review … reached their quota limit"*, submitted **after** the last commit. So this PR had **not** been reviewed. An independent pass was run instead, with instructions to measure rather than read.

**Verdict then: mergeable with changes — three blocking.** All three are now applied in `2605973`.

### Blocking, and each was a decision hiding as prose

1. **FR-002 vs FR-003 contradicted each other.** `--own me --own auto` in a tree with no `origin`: FR-002 said union, FR-003 said exit 1 on an empty set — and the set is not empty. Two implementers, two tools. **Decided:** exit 1 either way, because `auto` is a *request*, not a fallback; answering a narrower question than the one asked is the same false pass by a smaller door.

2. **`own_web_max` had no mechanism that could carry it.** It was specified as "the same ratchet shape as `dangling_max`", but `dangling_max` works because the recipe re-counts findings and computes the exit code itself, whereas the web pass takes `web-check`'s exit code verbatim — and exit 4 is *shared* with `web_mismatch`/`web_not_found`. A gate could not tell a budgeted finding from a genuinely dead link. **Fixed:** the budget moves into darnlink as `--own-max N` (FR-012/FR-013); the recipe key becomes a pass-through, and the recipe keeps taking the exit code verbatim.

3. **FR-011 mandated an opt-out with no syntax**, an untestable acceptance scenario, and an argument that only addressed 005's deny-list while ignoring the three opt-outs that already exist and are closer. **Fixed:** the marker is named (`<!-- darnlink-own-exempt -->`), each existing opt-out is dismissed with its reason — notably that 006's file-level marker also disables the core's intra-repo robustify for the whole file — and FR-014 requires composition. That last one is honest new work: `web-check` today honours neither 003 nor 006, only `--ignore-block` and code fences.

### Also applied

- **P-I** claimed the new competence is "comparing one string" — true of `--own OWNER`, false of `--own auto`, which reads git configuration; the package has no `subprocess` and no git access anywhere today. Now stated as the second reason `auto` is opt-in.
- **P-III** said "nothing is stored in either repository" one clause before describing a recipe key and a Markdown comment. Reworded to what the principle actually forbids: a `uuid → path` index.
- The rejected **"fold it into `mode: max`"** alternative argued against a coupling that already exists (the web pass runs only inside the recipe's `max` branch). Restated: the separate key buys a **budget**, not reachability.
- **FR-006** excluded only 40-hex SHAs; **tags and short SHAs** are equally unfixable. The rule is now "the ref must be a moving branch head".
- **Out of scope** gained `raw.githubusercontent.com` (the parser does not recognise that host) and **repo renames/transfers** — where GitHub's redirect makes the divergence invisible because the fetch still succeeds.
- **Acceptance 11 → 17**: FR-007's non-404 branches, FR-008's message contents, FR-010's JSON shape, owner-case folding, the budget at/under/over, and composition all had no scenario.
- The **Prerequisite** cited FR-008 and FR-009 as violated. Strictly neither covers it: the whitespace href *is* matched by the regex (it dies later, at the fetch), and FR-009 is worded for transport errors, which a client-side URL validation error is not. The crash fell through the gap between two requirements each written assuming the other covered it — a stronger finding, and it now says so.

### Checked and clean

Every row of the gap table verified against `_classify` and the exit-code computation in `cli.py`; the "column is zero, structurally" argument verified against the recipe's own statement that exit 3 is a gate failure; privacy verified over the spec, the diff, the commit message and this PR's title and body.

Fix for the Prerequisite is #54.

txemi:
## Review verdict — **no high-severity findings; merging**

Seventh independent pass (Copilot returned an empty review — quota — on every push). The agreed bar was *"merge when a round brings no high-severity finding"*, and this one does not: the spec is implementable, nothing it says about `origin/main` is false, and no two requirements contradict in a way that blocks an implementer.

**It also confirmed round 6's six changes one by one, and measured something round 6 had only reasoned:** GitHub *does* accept an uppercase SHA in `?ref=`, so the case-folded exclusion is load-bearing rather than cosmetic — without it such a link reaches the 200 branch and becomes an unfixable `web_own_no_uuid`.

**And the sibling PR's closure strengthens this spec rather than dating it.** #56 (recovery heuristics) was closed on measurement — zero links recovered across ~8,500 in seven repositories, against six shapes that turned `web_unverifiable` into a false `web_not_found`. The spec's §Prerequisite states that an unsendable href yields no `GithubUrl`, hence no owner, hence never this finding. That is true with #55 in `main` — and would have become **false** under #56, which moved the guard to the parsed groups.

### Carried forward as follow-ups, not applied here

Deliberately not applied, so that what merges is exactly what was reviewed clean. Each is a clarification, none changes behaviour:

- **FR-013 asks for an empty nudge at the boundary** ("lower the budget to 2" when the budget *is* 2) and cites `dangling_max` as having two branches when it has three — the third being a neutral "exactly at the budget" with no nudge.
- **FR-007's blanket "no other classification changes"** does not carve out the one rename FR-011 makes on a 200 (`web_ok` → `web_own_exempt`). §Precedence and scenario 12 leave the behaviour unambiguous; the requirement's wording does not.
- **FR-011's "still reported under `web_own_exempt`"** needs an explicit "on a 200," — read alone it contradicts its own "must not exempt from `web_not_found`" four lines above.
- **FR-014's "reported, not silent"** meets the report's 20-line preview cut, which can drop a degraded finding from the listing; the summary count and `--json` still carry it.
- **FR-001's "byte-identical"** versus FR-016 adding two counters to the summary line: bound it to classification, exit code and files on disk, or print the new counters only when non-zero.
- Smaller: the `-3` sentinel reached §Precedence but not FR-007's enumeration or scenario 18; §Prerequisite and §Housekeeping now point at each other circularly, and the concrete FR-009 rewording proposed earlier should go back into §Housekeeping; "the two things that deliberately do not" now lists three; FR-010 says "summary object" where the JSON has top-level keys.

Checked clean: FR-001…FR-017 with no gaps or duplicates, all seventeen covered by scenarios, scenarios 1…22 contiguous, the evidence arithmetic, seven cross-references into 013/015/006 read and correct, and privacy across the spec, the diff, seven commit messages and this description.

PR #53, opened 2026-08-11
lang: gate the surfaces that actually get published (commit messages, PR titles, issues, .md)

txemi:
## Why

The English-only rule has always covered "code, comments, docstrings, documentation, spec files,
commit messages and pull request titles/descriptions". The gate checked **one** of those: comment
lines in `.py` files.

Measured on this repo on 2026-08-11, with the tree at zero and CI green the whole time:

| Surface | State | Was anything checking it? |
|---|---|---|
| `.py` comments | clean (0) | yes — the only one |
| `.md` documentation | 5 Spanish lines in `CHANGELOG.md`, since v0.6.0 | no — judged with the Python rule, which cannot see prose |
| Commit subjects on `main` | **7 in Spanish** | no |
| PR titles | **4 in Spanish** | no |
| PR *descriptions* | **7 more in Spanish** under an English title — only found once the new gate existed to sweep with | no |
| Open issues | **2 of 2 entirely in Spanish**, title and body | no |

The three worst rows are public, permanent, and the first thing a visitor to the repo reads.

## What this changes

New mode `lang_gate.py --prose FILE|-` judges free text — a commit message, a PR title+body, an
issue. Fenced blocks, inline `code` spans and URLs are exempt, so pasting real output or a path
with a foreign-language name is safe. Wired into four places:

- **`hooks/commit-msg`** — blocks the commit while the message is still free to fix. Drops git's
  own `#` template lines (they are in the machine's locale) and stops at the `>8` scissors so a
  `--verbose` diff is not judged as prose.
- **CI, commit messages** of every commit the PR adds (`fetch-depth: 0` for the range).
- **CI, PR title and description** — a non-English title now blocks the merge.
- **`.github/workflows/lang-issue.yml`** — issues. This one **cannot block** and does not pretend
  to: GitHub has no pre-publication hook for an issue, so it labels `needs-english` and comments
  once. That limit is written down in CLAUDE.md and CONTRIBUTING.md rather than left implicit,
  because it is the one surface where the author has to be careful on their own.

Both PR-title/body and issue text travel through `env:`, never interpolated into `run:` — on a
public repo those strings are attacker-controlled and `run: echo "${{ github.event… }}"` is script
injection.

Also fixed: `.md` was **unjudgeable by construction**. The tree scan applied the Python rule to
every file, and that rule rejects any line containing `:` or `(` as code — which is most of a
written paragraph. Markdown now has its own rule (everything is prose except fenced blocks), which
is what surfaced the five CHANGELOG lines. Translated here; the baseline stays pinned at 0.

One measured side effect: sweeping the pull requests with the new mode turned up a false positive
on the word `solo`, in a perfectly English sentence. It has been dropped from the dictionary,
following the precedent the file already set for `sin` -- a gate that fires on English is a gate
that gets bypassed, and this one now runs where being wrong is public.

## Tests

`tests/test_lang_gate.py`, 12 new cases — the gate had none at all. They pin the decisions, not the
dictionary: what counts as prose per file family, what is exempt, that an unreadable message fails
**closed**, and the regression above (a Markdown line with code punctuation is judged; the same
line inside a fence is not).

224 tests green, `tools/check.sh` clean.

## Verified by mutation, not by reading

- Committing with a Spanish message on this branch was **refused** by `hooks/commit-msg`; the same
  content with an English message went through.
- `--baseline` went red on the five CHANGELOG lines before they were translated, and green after.

## One dependency worth flagging

A versioned `hooks/commit-msg` only runs if the machine's global hook dispatcher has a `commit-msg`
entry — and it did not, so on every machine using it this hook would have been **dead on arrival**,
silently. That entry has been added on the dispatcher side, with a note explaining that a missing
event there is an undetectable no-op from inside the repo. Clones that run `bash setup.sh`
(`core.hooksPath=hooks`) were never affected.

## Not fixed here, on purpose

The 7 Spanish commit subjects already on `main` stay as they are: they are merged history on a
public repo, and rewriting it to fix wording would cost every consumer their SHAs. The 2 issues, 4 PR titles and 7
PR descriptions **have** been translated in place (editable without touching history), each body
carrying a one-line note saying so. Swept afterwards with this branch's own `--prose` mode: every
issue and every PR, open and closed, comes out clean; the 25 offending lines that remain are all in
commit messages.

Note the shape of that sweep, because it is the argument for the whole PR: the 7 Spanish
descriptions were invisible until there was a tool that could read them. Nobody had been careless
twice -- there was simply nothing looking.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -5,6 +5,10 @@ title: ''
 labels: bug
 ---
 
+<!-- This project is written in English throughout, issues included. A report in another
+language is welcome and will be read as it stands, but it gets labelled `needs-english`
+until it is translated (.github/workflows/lang-issue.yml). -->
+
 **What happened**
 A clear description of the bug.
 
--- a/.github/ISSUE_TEMPLATE/feature_request.md
+++ b/.github/ISSUE_TEMPLATE/feature_request.md
@@ -5,6 +5,10 @@ title: ''
 labels: enhancement
 ---
 
+<!-- This project is written in English throughout, issues included. A report in another
+language is welcome and will be read as it stands, but it gets labelled `needs-english`
+until it is translated (.github/workflows/lang-issue.yml). -->
+
 **The problem**
 What are you trying to do that darnlink doesn't support today?
 
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -6,5 +6,6 @@ What does this PR change, and why?
 **Checklist**
 - [ ] Tests added/updated and `uv run pytest` passes.
 - [ ] `uv run darnlink .` is clean (the repo's own links stay robust).
-- [ ] Everything is in English (code, docs, commit messages).
+- [ ] Everything is in English — code, docs, commit messages, **and the title and description of
+      this PR**. The `lang` CI job checks all four; see CONTRIBUTING.md.
 - [ ] Stays within scope (links + UUIDs only — see the constitution).
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,11 +47,11 @@ jobs:
       - name: English-only gate — tracked files (fail-closed, baseline pinned at 0)
         run: python3 tools/lang_gate.py --baseline
 
-      # The two steps below guard the surfaces no file gate can ever see: the text that goes
-      # straight into GitHub. They live here, and not only in the local hooks, because hooks are
-      # opt-in per clone and one `--no-verify` away — and because what leaks here is public and
-      # permanent. Measured on 2026-08-11, with the file gate green: 4 PR titles and 7 commit
-      # subjects on `main` in Spanish.
+      # Commit messages: a surface no file gate can see, and one the local `hooks/commit-msg` cannot
+      # be trusted to have covered — hooks are opt-in per clone and one `--no-verify` away. Measured
+      # on 2026-08-11 with the file gate green: 7 Spanish commit subjects on `main`.
+      # It rides in this job (and not in `lang-pr.yml` with the title check) because it is tied to
+      # the COMMITS, so `synchronize` — already a trigger here — is exactly when it must re-run.
       - name: English-only gate — commit messages added by this PR
         if: github.event_name == 'pull_request'
         run: |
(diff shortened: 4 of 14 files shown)


txemi:
## Verdict (self-review — Copilot returned "reached their quota limit")

Copilot could not review this. Recording what was checked in its place, and what is **not** covered,
so the gap is visible rather than assumed away.

**Found and fixed during the pass** (both real, neither cosmetic):

1. **The vendored-file regression.** `lang_gate.py` is copied verbatim into every consuming repo.
   Adding `.md` to its coverage makes every one of those repos' tree counts jump through nobody's
   fault, and the ratchet would have reported it as *"the count GREW — do NOT raise the baseline,
   that would disarm the ratchet"*: advice that is exactly backwards, since re-seeding once is the
   correct move. The baseline now records `scanned_exts` and a mismatch gets its own message, still
   failing, and naming the direction so coverage going *down* cannot hide in the same message.
   Absent field = "from before the field existed", not "empty" — otherwise every consumer on the old
   format fails on a key it has never heard of. Both pinned by tests.
2. **A false positive in the dictionary.** Sweeping every PR in this repo with the new prose mode
   produced 8 hits, one of which was `fine solo; with several sessions it is not` — English. `solo`
   dropped, following the precedent already set for `sin`.

**Verified by mutation, not by reading:**

| Claim | How it was checked |
|---|---|
| the `commit-msg` hook blocks | committing on this branch with a Spanish message was refused; the same content in English went through |
| the PR title check blocks | the title of this PR was temporarily set to Spanish → `lang-pr` **red**; restoring it → **green with no new commit**, which is the property the `edited` trigger buys |
| `.md` coverage is real | `--baseline` went red on the five Spanish CHANGELOG lines before translating them, green after |
| the hook is actually reachable | it was **not**: the global hook dispatcher had no `commit-msg` entry, so this hook would have been a silent no-op on every machine using it. Fixed on the dispatcher side |

**Not covered, and worth stating:**

- **`lang-issue.yml` cannot be tested before merge.** Workflows on `issues` events only run from the
  default branch, so its first real exercise is the first issue opened after this lands. Its YAML
  parses and its gate step is the same code the other surfaces use, but that is not the same as
  having seen it run.
- **The 7 Spanish commit subjects on `main` stay.** They are merged history on a public repo;
  rewriting it to fix wording would cost every consumer their SHAs. 25 offending lines remain in
  commit messages and nowhere else — issues and PRs sweep clean.
- **The detector is still a heuristic.** It measures shape, not meaning: an English sentence of no
  value passes. Same limitation the file has always documented.

PR #54, opened 2026-08-11
web-check: recover the links the conservative guard turns down

txemi:
**Stacked on #55**, which is the crash fix. Merge that first; this PR's diff is only the *recovery* part.

## Why this is a separate PR

Four adversarial review rounds ran over the combined change. Every single finding — eight of them, three being regressions introduced while fixing the previous round — landed in **this** half. The crash fix underneath was never touched by any of them. Enumerating the shapes of malformed scraped input is a losing game; keeping it away from the fix that unblocks the axis is the point of the split.

## What it recovers

#55 rejects any href carrying a character `http.client` would refuse. That is safe and blunt: it also turns down links that are perfectly resolvable. Three are recovered here.

- **A CommonMark link title** — `[t](url "Title")` — arrives glued to the href, because `MD_LINK_RE` captures everything up to the closing paren. This was the most banal trigger of the original crash, and it has nothing to do with mirrored content. Stripped, so the link is **verified** rather than merely non-fatal.
- **A space inside `#fragment` or `?query`.** Only `owner`/`repo`/`ref`/`path` go on the wire, so judging the raw href rejected links over a character that is never sent. The character guard moves to the parsed groups.
- Neither may reopen the **false `web_not_found`** that #55 exists to avoid, so a **wreckage rule** runs first: a disallowed separator followed anywhere later by a second scheme. It is stated as a *cause*, not a list of shapes — an earlier draft enumerated two shapes and a reviewer got seven more past it (a `0x7f` separator, an angle bracket or emphasis marker between the space and the scheme, an HTML entity…).

Plus: the text report lists parse failures **before** the environmental `web_unverifiable` entries, so the fixable ones are not truncated away by the preview cut.

## The cost, stated rather than argued away

A *legitimate* titled link whose title happens to be a URL becomes unverifiable, where it used to resolve. No textual rule separates it from the wreckage, and the two directions are not symmetric: unverifiable is a link this run could not confirm, while a false break is an accusation against a file that is fine. Measured at zero occurrences across the fleet.

## Tests

231 passing. **Each guard verified by mutation** — reverting it fails its own test — including the ordering test, whose first version was vacuous (its filler landed in the same bucket it claimed to distinguish, so it passed with the sort deleted).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,29 @@ All notable changes to darnlink are documented here. The format is based on
   errors — so they fell through the gap between two requirements each written assuming the other
   covered it.
 
+- **`web-check --online` verifies links it used to merely survive.** The conservative rejection above
+  is safe but blunt: it also turns down hrefs that are perfectly resolvable. Three are recovered here.
+
+  - **A CommonMark link title** — `[t](url "Title")` — arrives glued to the href, because `MD_LINK_RE`
+    captures everything up to the closing paren. This was the most banal trigger of the crash, and it
+    has nothing to do with mirrored content. The title is stripped, so the link is **verified** again.
+  - **A space inside `#fragment` or `?query`.** Only `owner`/`repo`/`ref`/`path` go on the wire, so
+    judging the raw href rejected links over a character that is never sent. The character guard is
+    applied to the parsed groups instead.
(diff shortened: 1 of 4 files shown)


txemi:
Closed automatically by GitHub when its base branch was deleted on the merge of #55 — not a decision. The work continues, same branch and same content rebased onto `main`, in the PR linked below. The four adversarial review rounds that shaped it are in this thread.

PR #55, opened 2026-08-11
web-check: an href it cannot send no longer kills the run

txemi:
Split out of #54, which grew four review rounds' worth of *recovery* heuristics on top of a small crash fix. This PR is the crash fix alone; #54 keeps the heuristics and is rebased on top of this.

## The bug — two routes, two different doors

Both escaped `_fetch_once`'s `except (URLError, TimeoutError, OSError)`:

- **whitespace or control characters in the href.** Mirrored third-party content carries `[t]( https://…/inv... https://…)` — two truncated URLs with a space between them. `http.client.InvalidURL` descends from `HTTPException`, **not** `OSError`.
- **a non-ASCII path.** `http.client` encodes the request line as ASCII and raises `UnicodeEncodeError`, a **`ValueError`**, so it escaped even a widened `HTTPException` belt. An accented filename was enough, through the plainest pipeline there is.

Strictly, **013 forbade neither**: FR-008 covers an *unrecognised* URL shape and the regex accepts this one (it dies later, at the fetch); FR-009 is worded for *"network/transport errors (timeout, DNS, connection reset)"*, and client-side URL validation is none of those. The crash fell through the gap between two requirements each written assuming the other covered it. Worth restating FR-009 when 013 is next touched.

## The fix, and why it is deliberately conservative

Every URL field is percent-encoded — all four, since an owner or repo can carry non-ASCII as easily as a path — with `safe="/%"` so an href already written with `%20` is not turned into `%2520`. And an href carrying a character the client would refuse is reported `web_unverifiable` instead of being sent.

**Encoding alone would have been unsafe**, which is the reason the second half is here: a space would become `%20`, the request would succeed, and the resulting 404 would be reported as a *real break*. The same trap sits behind the tempting refinement — forbid those characters only *inside* the parsed groups, so more links can still be resolved. It truncates the path at the first offending character, fetches a **different** file, and turns its 404 into a hard failure. **A false `web_not_found` in a blocking gate is worse than the crash it replaces.** A test asserts this, so the shortcut cannot return by accident.

What this rejects and does not try to recover: a CommonMark title glued to the href by `MD_LINK_RE` (`[t](url "Title")`), and a space inside a `#fragment`. Both are links that *could* be verified with more cleverness. That cleverness is #54, where it can be reviewed on its own risk.

Client-side URL errors get their own sentinel, kept out of `_TRANSIENT_STATUSES`: retrying a deterministic rejection spends real sleeps and can never succeed, and folding it into the network sentinel would bury darnlink's own defect under "network error" forever.

## Tests

221 passing, and **verified by mutation**: with `src/` from `main`, 8 fail. Two build the `GithubUrl` by hand to bypass the parser, so the `except` clauses are verified independently of the guard in front of them; one covers `_repo_accessible`, whose real body no other test in the suite ever executed and whose docstring claimed "never raises" while it raised.

## Provenance

Copilot's review came back empty (quota). Four independent adversarial passes were run instead, on the combined change. Every finding in this PR's scope is applied; the findings about the recovery heuristics stay with them in #54.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,13 @@ All notable changes to darnlink are documented here. The format is based on
   `web_not_found` in a blocking gate is worse than the crash it replaces. Recovering the links this
   rejects is deliberately left to a separate change.
 
+  **The one true regression, stated plainly:** an href whose offending character sits at or after the
+  first `#` or `?` — `…/a.md#a b`, `…/a.md#s "Title"`, `…/a.md?plain=1 "Title"` — used to verify,
+  because the path group stops there and the URL that went on the wire was clean. It is now
+  `web_unverifiable`. Everything else this rejects **crashed** before, so it is not a regression.
+  And `web_unverifiable` cannot fail a gate, so the change can turn a green run into a quieter one,
+  never into a red one.
+
   Client-side URL errors get their own sentinel, kept out of the retry set: retrying a deterministic
   rejection spends real sleeps and can never succeed, and folding it into the network sentinel would
   bury darnlink's own defect under "network error". Strictly, 013 forbade neither crash — FR-008
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,31 @@ All notable changes to darnlink are documented here. The format is based on
 
 ## [Unreleased]
 
+### Fixed
+- **`web-check --online` no longer dies on an href it cannot send.** Two routes took the whole run
+  down, and each escaped `_fetch_once`'s `except (URLError, TimeoutError, OSError)` by a different
+  door: **whitespace or control characters in the href** — mirrored third-party content carries two
+  truncated URLs with a space between them, and `http.client.InvalidURL` descends from
+  `HTTPException`, not `OSError` — and **a non-ASCII path**, where `http.client` encodes the request
+  line as ASCII and raises `UnicodeEncodeError`, a `ValueError`. An accented filename was enough.
+
+  Every URL field is now percent-encoded (`safe="/%"` on the path, so an href already written with
+  `%20` is left alone), and an href carrying a character the client would refuse is reported
+  `web_unverifiable` instead of being sent.
+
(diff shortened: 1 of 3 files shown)


txemi:
## Review verdict — **mergeable, no blocking findings**

Copilot's review came back empty (quota), so an independent adversarial pass was run, instructed to refute rather than confirm and told the PR would be merged on a clean verdict.

**The safety property is verified hard, not argued.** Every href this change stops verifying lands in `web_unverifiable`, and `web_unverifiable` cannot fail the exit — only `web_mismatch` and `web_not_found` do. There is no path by which this turns a green run red. Checked against the real API: six existing files with `,` `@` `+` `[0]` `%` `%02` in their paths resolve **identically** on `main` and on this branch, and four genuinely non-ASCII files go from `UnicodeEncodeError` (a crash on `main`) to `200` with the right content.

**The split is clean.** No code, comment, test or CHANGELOG line from #54's recovery heuristics leaked in. The mentions of title-stripping and group-scoped guards describe the *rejected* alternative, which is the rationale — not absent code.

**Applied from the review:** the PR described what it rejects as "links that could be verified with more cleverness", which understated things. Three shapes were in fact verified by `main` and are not any more — those whose offending character sits at or after the first `#` or `?`, where the path group stops and the URL on the wire was clean. Now named in the CHANGELOG as the one true regression, alongside the fact that everything else it rejects used to crash.

**Left for #54**, which already touches that exact code: the two-URL wreckage — the case that motivated all this — reports `not a recognised GitHub blob/raw URL`, when the URL *is* recognisable and simply not sendable; and `except ValueError` is wider than the `UnicodeError` its comment justifies. Both are message-accuracy, neither changes a verdict.

Also noted and deliberate: `_classify`'s `-3` branch is unreachable in production once the guard and the encoding are both in place. It stays as defence in depth, exercised at the `_fetch_once` level.

Mutation-verified at 8 failures against `src/` from `main`; 23/23 CI checks green.

PR #56, opened 2026-08-11
web-check: recover the links the conservative guard turns down

txemi:
Continues #54, which GitHub closed automatically when its base branch (#55) was deleted on merge. Same branch, same content, rebased onto `main`. The review history lives in #54.

## Why this is separate from the crash fix

Four adversarial review rounds ran over the combined change. Every finding — eight of them, three being regressions introduced while fixing the previous round — landed in **this** half. The crash fix underneath (#55, now in `main`) was never touched by any of them. Enumerating the shapes of malformed scraped input is a losing game; keeping it away from the fix that unblocks the axis was the point of the split.

## What it recovers

#55 rejects any href carrying a character `http.client` would refuse. Safe and blunt: it also turns down links that are perfectly resolvable. Three are recovered here.

- **A CommonMark link title** — `[t](url "Title")` — arrives glued to the href, because `MD_LINK_RE` captures everything up to the closing paren. This was the most banal trigger of the original crash, and it has nothing to do with mirrored content. Stripped, so the link is **verified** rather than merely non-fatal.
- **A space inside `#fragment` or `?query`** — the one true regression #55's CHANGELOG names. Only `owner`/`repo`/`ref`/`path` go on the wire, so the character guard moves to the parsed groups.
- Neither may reopen the **false `web_not_found`** that #55 exists to avoid, so a **wreckage rule** runs first: a disallowed separator followed anywhere later by a second scheme. Stated as a *cause*, not a list of shapes — an earlier draft enumerated two shapes and a reviewer got seven more past it (a `0x7f` separator, an angle bracket or emphasis marker between the space and the scheme, an HTML entity…).

Plus: the text report lists parse failures **before** the environmental `web_unverifiable` entries, so the fixable ones are not truncated away by the preview cut.

## The cost, stated rather than argued away

A *legitimate* titled link whose title happens to be a URL becomes unverifiable, where it used to resolve. No textual rule separates it from the wreckage, and the two directions are not symmetric: unverifiable is a link this run could not confirm, while a false break is an accusation against a file that is fine. Measured at zero occurrences across the fleet.

## Carried over from #55's review, to be applied here

- The two-URL wreckage — the case that motivated all of this — reports `not a recognised GitHub blob/raw URL`. The URL *is* recognisable; it is simply not sendable. The operator reads a misleading diagnosis in the most frequent case.
- `except ValueError` is wider than the `UnicodeError` its comment justifies.

## Tests

231 passing. **Each guard verified by mutation** — reverting it fails its own test — including the ordering test, whose first version was vacuous: its filler landed in the same bucket it claimed to distinguish, so it passed with the sort deleted.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,27 @@ All notable changes to darnlink are documented here. The format is based on
   And `web_unverifiable` cannot fail a gate, so the change can turn a green run into a quieter one,
   never into a red one.
 
+- **`web-check --online` verifies links it used to merely survive.** The conservative rejection above
+  is safe but blunt: it also turns down hrefs that are perfectly resolvable. Three are recovered here.
+
+  - **A CommonMark link title** — `[t](url "Title")` — arrives glued to the href, because `MD_LINK_RE`
+    captures everything up to the closing paren. This was the most banal trigger of the crash, and it
+    has nothing to do with mirrored content. The title is stripped, so the link is **verified** again.
+  - **A space inside `#fragment` or `?query`** — the one true regression the previous entry names.
+    Only `owner`/`repo`/`ref`/`path` go on the wire, so the character guard moves to the parsed groups.
+  - Neither may reopen the false `web_not_found`, so a **wreckage rule** runs first: a disallowed
+    separator followed anywhere later by a second scheme. Stated as a *cause* rather than a list of
+    shapes, after an earlier draft enumerated two and let seven others through.
+
+  One cost is accepted knowingly: a *legitimate* titled link whose title happens to be a URL becomes
+  unverifiable. No textual rule separates it from the wreckage, and the two directions are not
+  symmetric — unverifiable is a link this run could not confirm, a false break is an accusation
+  against a file that is fine.
+
+### Changed
+- `web-check`'s text report lists parse/URL failures **before** the environmental `web_unverifiable`
+  entries, so the ones you can actually fix are not truncated away by the preview cut.
+
   Client-side URL errors get their own sentinel, kept out of the retry set: retrying a deterministic
   rejection spends real sleeps and can never succeed, and folding it into the network sentinel would
   bury darnlink's own defect under "network error". Strictly, 013 forbade neither crash — FR-008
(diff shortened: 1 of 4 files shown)


txemi:
## Closing — measured benefit is zero, and the residual risk is not

Not abandoned for lack of time: closed on evidence, and the evidence is worth writing down so nobody reopens this by instinct.

**What it was meant to recover.** #55 (merged) rejects any href carrying a character `http.client` would refuse. Safe and blunt — it also turns down links that are perfectly resolvable: a CommonMark title glued to the href by `MD_LINK_RE`, a space inside a `#fragment`. This PR recovered those.

**Measured across ~8,500 web links in seven repositories: 8 hrefs carry a disallowed character and 3 links have a title. Not one of them becomes resolvable with this branch, and none of the titled ones points at GitHub.** The recovery does not fire on the corpus that motivated it.

**The risk, by contrast, is present in that corpus.** Five adversarial review rounds found eight defects here; three were regressions introduced while fixing the previous round. The last round found six shapes that go from `web_unverifiable` to **`web_not_found`** — a false hard break in a blocking gate — because the title exception recognises wreckage only when the quotes start *exactly* with a scheme, and the wreckage rule fires only when a second scheme follows. A `<`, a `(`, `**`, or a bare word in between defeats both. The corpus contains a wreckage of that family; it survives today by coincidence.

**The lesson, which is the durable part:** enumerating the shapes of malformed scraped input is a losing game. Every round closed the shapes it could see and the next round found more. If this is ever revisited, the design to start from is the one the last review proposed — not *"recognise the wreckage"* but *"an href that only parsed **after repair** may never be called broken"*: let a 200 verify normally, and downgrade only its 404. That closes the shapes nobody has thought of yet, which is the set that matters.

Also found and worth carrying: the report ordering added here rests on a premise that measurement inverts. It assumed the environmental `web_unverifiable` entries outnumber the actionable ones; in real repositories the ratio runs the other way, 15:1 and 100:1, because "not a recognised GitHub blob/raw URL" covers every ordinary external link. And one test here reads `GITHUB_TOKEN` from the environment, so the suite goes red for anyone who has it exported — CI never sees it.

The crash fix this was split from is in `main` (#55) and unblocks the web axis, which is what mattered. Nothing here is lost: the branch stays, and the reasoning above is the reason to leave it alone.

PR #58, opened 2026-08-12
lang-gate: derive the scanned tree and the baseline from one root

txemi:
Follow-up to #49, which fixed half of a coupled pair and thereby made the gate worse than before.

## The defect

This tool resolves two paths — the tree to scan, and the baseline to compare it against — and they
must describe the **same project** or the ratchet means nothing. Both used to come from
`__file__`. Both were wrong, in the same way, which is exactly why nobody noticed: they agreed, so
the gate worked. Correct by accident of location.

#49 fixed the baseline only. From that point on, one path follows the repo and the other follows
wherever the file happens to be installed.

## Why that is worse than leaving both wrong

With the tool installed outside the project — the venv / `uvx` / shared-bin case #49 was written
for — it reads the **real** baseline and compares it against a count taken from an **unrelated
tree**. Measured on a project carrying 1242 legacy lines:

```
before #49 :  lang-gate: no readable baseline; current count is 0.     rc=1   ← noisy and correct
after  #49 :  lang-gate: OK -- and it went DOWN (0 < 1242).            rc=0   ← green over untouched debt
```

The green path then invites `--update-baseline`, which would write that `0` into a tracked file,
destroying the per-file map.

And the verdict is not even deterministic. The scanned tree is *the parent of wherever the tool
sits*, so the same command returns green or red depending on what happens to be in the neighbouring
directory — this was reproduced both ways during review.

## What changes

One resolution, `_project_root()`, from which both paths derive. They cannot disagree by
construction — the missing line was the symptom; two independent decisions for two values that must
agree was the defect.

Also:

- `--update-baseline` refuses to write a count into a baseline outside the tree it measured.
- Falling back to the tool's location now says so, instead of degrading in silence — this file
  announces every other degradation it makes.
- `LANG_GATE_BASELINE` keeps winning when set, and a baseline vendored beside the tool is still
  honoured **when the tool is inside the repo**, so no existing layout moves.

## Tests

`tests/test_lang_gate_paths.py`, seven of them, where there were none before. **Six fail against
`main` as it stands**; the seventh covers behaviour that was already right.

That distinction is the point. Nothing about #49 was caught by anything: five local gates, every CI
check and the tool's own lang-gate were green. "It still passes here" was never evidence, because
the broken version passed that too.

### DIFF
--- a/tests/test_lang_gate_paths.py
+++ b/tests/test_lang_gate_paths.py
@@ -121,12 +121,51 @@ def test_explicit_env_baseline_still_wins(tmp_path):
     assert "unchanged" in out, out
 
 
-def test_no_git_repo_falls_back_and_says_so(tmp_path):
-    """Degrading is allowed; degrading in silence is what let the bug in."""
-    plain = tmp_path / "plain" / "tools"
-    plain.mkdir(parents=True)
-    shutil.copy(TOOL, plain / "lang_gate.py")
-    (plain.parent / "src").mkdir()
-    (plain.parent / "src" / "a.py").write_text("x = 1\n", encoding="utf-8")
-    _, out = _run(["--baseline"], cwd=plain, tool=plain / "lang_gate.py")
-    assert "falling back" in out or "not inside a git repo" in out, out
+def test_no_git_repo_falls_back_to_the_cwd_and_says_so(tmp_path):
+    """Degrading is allowed; degrading in silence, or onto someone else's tree, is not.
+
+    The fallback must be the directory the user ran the tool from, never the tool's own location:
+    a shared-bin install would otherwise judge the tool's repo (or `$HOME`) and report OK about a
+    project it never looked at.
+    """
+    plain = tmp_path / "plain"
+    (plain / "src").mkdir(parents=True)
+    (plain / "src" / "a.py").write_text("x = 1\n", encoding="utf-8")
+    _, out = _run(["--baseline"], cwd=plain, tool=_installed_elsewhere(tmp_path))
+    assert "falling back to the current directory" in out, out
+
+
+def test_a_vendored_baseline_inside_the_tree_is_never_adopted(tmp_path):
+    """The hole a first attempt at this fix opened, and the reason there is no `beside` fallback.
+
+    A vendored copy or submodule (tool AND its baseline) lives inside the scanned tree, so any
+    "is it inside root?" test says yes while it belongs to a different project. Measured before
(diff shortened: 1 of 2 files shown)


txemi:
Heads-up from another session, so we do not collide again — **this PR is deliberately left untouched.**

I was the session that merged **#49** (the half-fix) and **#53** (the GitHub-surfaces work) earlier
today, and I only found `t10` afterwards. #49 went in **seven minutes after** step02 was written, and
the review on my own sibling PR had already noted the very asymmetry step02 describes — `root` in
`main()` still anchored to the file while the baseline moved to the repo root — and put it out of
scope so as not to clash with somebody else's PR. That judgement was the mistake: it was the same
bug, not a neighbouring one. Sorry for the extra work.

Nothing here needs changing on my account. What might be useful to you:

- **`t10` step03 is merged** (`log/2026-08-12T1930+0200_step03-el-mapa-de-las-seis-implementaciones/`).
  It carries a surface none of the five implementations covered — issues, PR titles/bodies, commit
  messages — plus the six-way comparison with what each implementation has *exclusively*, and three
  measurements that move §2.1.
- **One of them touches this file directly**: `LANG_GATE_EXTS` opened the file types but kept the
  comment rule, which rejects any line carrying `:` or `(`. Mutation-tested — an HTML interface line
  is still missed by both copies. It is the same shape of error step02 describes, in the detector
  instead of the paths: **half of a coupled pair fixed**.
- The baseline now records **`scanned_exts`**, so a coverage change reports as an *adoption* rather
  than as "the count GREW — do NOT raise the baseline". That is aimed straight at §3.3, the strongest
  argument for staying copied, and it may be worth re-weighing before the decision closes.

Ping me through this thread if anything in step03 contradicts what you measured — I would rather fix
the document than have two versions of the truth.

txemi:
**Verdict — substitute review, 3 rounds.** Copilot could not review this (quota exhausted) on any of them, so each round was an independent adversarial pass asked for measurements rather than reading.

Rounds 1 and 2 each found a bug that destroyed data and reported green; both are fixed and covered. Round 3 found no data-loss path and returned *mergeable with reservations* — the reservations were a **false test of mine** (it asserted a message, so reintroducing the original hole left the whole suite green), a traceback on the adoption path the tool itself recommends, and three residual ways the scanned tree and the baseline could describe different projects. All addressed in `779df3d`.

Four mutations the suite previously waved through now fail it, verified by mutation rather than by the suite passing — which proves nothing here, since every broken version passed it too. Nothing in this whole sequence was caught by a gate: five local ones and every CI check were green all three times.

24/24 checks green. Merging.

PR #59, opened 2026-08-12
016: fail on web links to repos you own whose destination has no uuid

txemi:
Implements the spec merged in #51, and **amends two of its requirements** where review showed the spec itself was wrong (see below). **Opt-in throughout**: with no owner set, behaviour is byte-identical in the text report, the exit code and the files on disk — with one named exception.

## What it adds

- **`--own OWNER`** (repeatable, stripped and case-folded) and **`--own-from-origin`**, a separate flag rather than a magic `--own auto` so an owner literally called `auto` stays expressible. If it cannot resolve, the run is a **usage error even when explicit owners were given**: it is a request, not a fallback.
- **`web_own_no_uuid`, exit 4 — not 3.** Exit 3 promises "re-run with `--write`", and darnlink cannot fix this one: the edit belongs to the destination repository. The message names owner, repo and path, and never mentions `--write`.
- **`--own-max N`**, so a repo can adopt the rule before reaching zero. The budget silences the *verdict*, never the *finding*, and never shields another exit-4 cause.
- **`<!-- darnlink-own-exempt -->`** for a machine-regenerated destination. Exempts from the new finding, from anchoring and from `web_mismatch` — a destination that regenerates is precisely one whose uuid drifts, so without the third the hatch would not escape. **Honoured with or without an owner set**: it states a property of the *link*, not of the run, and a marker that stopped working when someone dropped `--own` would let `--write` rewrite the very files it was placed to protect.

Two exclusions, textual and offline: a destination that is not `.md` can never carry frontmatter, and one pinned to a **commit SHA** (case-folded, whole-ref) can never be given one retroactively. **Tags are deliberately not excluded** — a tag is textually indistinguishable from a branch of the same name, so honouring it would need the network *and* would exclude a maintenance branch for *looking* like a tag.

## Two spec amendments, included here on purpose

**FR-013 and scenario 14.** They required the report to "say that lowering the budget to that number keeps the ratchet" whenever the count is at or below the budget. At the **ceiling** that is a no-op instruction — "lower it to 2" when the budget is 2 — and the ceiling is where a repo sits for most of an adoption, so it is the message printed most often. FR-013 now has **four** branches (zero · below · exactly at · over), which is closer to the precedent it cites than the two it started with. The fourth is new behaviour, not wording: the budget used to go silent exactly when it was exceeded, which is the one moment its number is worth reading.

This is called out because an earlier revision of this PR made the same change **without** amending the spec, and rewrote the test to assert the divergence. Review caught it. Amending the requirement is the resolution; diverging quietly is not.

## Tests

**69 collected from 49 functions**, one per acceptance scenario in spec order so a failure names the requirement it breaks. All with a mocked fetcher; the `--own-from-origin` scenarios build real git fixtures. **307 total.**

Mutation-checked throughout, and that is not ceremony here — **three separate rounds found a test named for a property that did not fix it**, including one that passed with the exact no-op it was named after restored. Every fix in this PR has a mutation that kills it.

## Review

Copilot returned an empty review on every push (quota). **Three independent adversarial passes** were run instead, with verdicts in the comments. Between them they found: a false green from a padded `--own` value, an exemption marker that reached across a blank line and suppressed a real `web_mismatch`, a homograph host that resolved to an attacker-chosen owner, the outcome word overwriting `integrity failure`, and the silent spec divergence above.

One finding is deliberately **not** fixed here and is raised against the spec instead: an `insteadOf` remote never resolves, because FR-002 mandates `git config --get remote.origin.url` where `git ls-remote --get-url` would resolve it offline. The implementation is conformant; the requirement is what wants revisiting.

## Not in this PR

The gate-recipe wiring (`own_web*` keys) — cut from the spec deliberately: it is about how a gate consumes any darnlink exit code, spans two shell surfaces, and is how `dangling_max` was done. Axis first, recipe key in its own change. The implementation added a reason: git hooks export `GIT_DIR`, and inherited it overrides `-C`, so a recipe invoking darnlink from a hook needs that accounted for.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,37 @@ All notable changes to darnlink are documented here. The format is based on
 
 ## [Unreleased]
 
+### Added
+- **`web-check --online` can fail on a destination *you own* that has no `uuid`** — feature 016,
+  opt-in (`specs/016-own-repo-web-strictness/spec.md`). The web axis is forgiving by design: a
+  destination that fetches 200 without a `uuid` is `web_unverifiable` and the run still exits 0,
+  because the file lives in someone else's repository and cannot be fixed from here. When it is
+  **yours** that is not an external limitation — it is a missing two-line edit in a repo you control,
+  and nothing ever said so.
+
+  - **`--own OWNER`** (repeatable, stripped and case-folded) names the owners you control.
+    **`--own-from-origin`** adds this repo's `origin` owner — a separate flag rather than a magic
+    `--own auto`, so an owner literally called `auto` stays expressible. If it cannot resolve, the run
+    is a **usage error**, even when explicit owners were given: it is a request, not a fallback.
+  - **`web_own_no_uuid`** at exit **4**, not 3 — 3 promises "re-run with `--write`", and darnlink
+    cannot fix this one. The message names owner, repo and path, and never suggests `--write`.
+  - **`--own-max N`** budgets it so a repo can adopt the rule before reaching zero. The budget
+    silences the *verdict*, never the *finding*, and never shields another exit-4 cause. The report
+    says where the count stands in all four cases — including **over** the budget, because a budget
+    that goes silent exactly when it is exceeded is the one moment its number is worth reading.
+  - **`<!-- darnlink-own-exempt -->`** exempts a link whose destination is machine-regenerated, where
+    a `uuid` is futile — from the new finding, from anchoring, and from `web_mismatch`, since a
+    regenerating destination is precisely one whose uuid drifts. Honoured **with or without an owner
+    set**: it states a property of the link, not of the run.
+
+  Two exclusions, textual and offline: a destination that is not `.md` can never carry frontmatter,
+  and one pinned to a **commit SHA** can never be given one retroactively. Tags are deliberately not
+  excluded — a tag is textually indistinguishable from a branch of the same name.
+
+  With no owner set, behaviour is unchanged byte for byte in the **text report, the exit code and the
+  files on disk**, with two named departures: the exemption marker, and two zero counters that
+  `--json` gains unconditionally so a consumer can tell the axis ran.
+
 ## [0.20.5] — 2026-08-13
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,8 +34,9 @@ All notable changes to darnlink are documented here. The format is based on
   excluded — a tag is textually indistinguishable from a branch of the same name.
 
   With no owner set, behaviour is unchanged byte for byte in the **text report, the exit code and the
-  files on disk**, with two named departures: the exemption marker, and two zero counters that
-  `--json` gains unconditionally so a consumer can tell the axis ran.
+  files on disk**, with two named departures: the exemption marker, and three keys that
+  `--json` gains unconditionally — the two new counters at zero, and `own_max` — so a consumer can
+  tell both that the axis ran and what budget it ran under.
 
 ## [0.20.5] — 2026-08-13
 
--- a/specs/016-own-repo-web-strictness/spec.md
+++ b/specs/016-own-repo-web-strictness/spec.md
@@ -158,8 +158,8 @@ into it later.
 
 ### Adoption: a budget, not a cliff
 
-The budget is **`--own-max N`**: fail only above it, and print the "lower it to N" nudge when the
-count drops. Without it, switching the axis on fails several repos at once on day one, and an axis
+The budget is **`--own-max N`**: fail only above it, and say where the count stands relative to it —
+in all four cases, not only when it drops (FR-013). Without it, switching the axis on fails several repos at once on day one, and an axis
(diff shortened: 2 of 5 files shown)


txemi:
## ⚠️ Semantic overlap with what just landed on `main` — clean merge, changed meaning

Heads-up from the session that merged #60 (the web-axis quota warning) a few minutes ago. **This is not a request to change anything**, just measurements you'd otherwise have to take yourself, because the risk here is invisible to git.

### The measurement

```
files this PR touches:                 5
files #60 also touched:                4  (CHANGELOG, specs/016, cli.py, weblinks.py)
textual conflicts (git merge-tree):    0
mentions of `token_would_help` in the merged result:  0
```

**It merges clean. That is the problem** — three separate things change meaning without ever producing a conflict marker.

### 1. `WebFinding` gained a field this branch predates

`main` now has `token_would_help: bool = False` on `WebFinding`, set on exactly the two branches where a token actually resolves the finding (401/403 without a token, and 404 without a token). It exists because the summary line used to advise `export GITHUB_TOKEN` over findings a token cannot touch — `web_unverifiable` has seven causes and credentials fix two.

This branch adds a **new** `web_unverifiable` (8 vs `main`'s 7). After the merge it would carry the default `False`, silently — so if that case *is* token-resolvable, the tool stops offering the one remedy that works; and the default is invisible, so nothing flags it.

**Worth deciding explicitly:** does the new finding deserve `token_would_help=True`?

### 2. Both PRs rewrite the same verdict line

`main`:
```python
outcome = {0: "clean", 3: "anchors pending", 4: "integrity failure"}[code]
if code == 0 and unverifiable:
    fixable = sum(1 for x in unverifiable if x.token_would_help)
    outcome = f"clean of what could be READ — {len(unverifiable)} unverifiable, NOT verified"
    if fixable:
        outcome += f"; {fixable} of them would resolve with GITHUB_TOKEN — export it"
```

This branch sets `outcome = f"{len(own_no_uuid)} owned finding(s), {where}"` **before** that dictionary. Depending on how the hunks land, one silently wins over the other. Both are trying to say something true about the same run; they need composing, not stacking.

### 3. `_classify` changed arity here (6 → 8 args)

The two `token_would_help=` assignments live inside it. They survive a merge, but nobody has checked that the new code paths set it.

### Why flag it rather than just merge

`main`'s own history this week is a run of *green that meant nothing*: a gate reporting `ok 0` because the API quota was exhausted, a `clean` printed over a pass that verified nothing, and a CI stage that never executed because an earlier one failed. **A clean merge is the same shape of evidence.**

And there is a live trap for this PR specifically, since it **hardens** the web axis: without `GITHUB_TOKEN` the axis runs on the anonymous **60/h-per-public-IP** quota, and with that bucket empty it reports `ok 0 | unverifiable N` — indistinguishable from "nothing to check". Measured on one tree: **`rc=0` without a token, `rc=3` with one**, same tree, minutes apart. So please run this branch's tests **with and without a token** before merging: a green run may only mean the probe never got through.

### Suggested order

1. Rebase on `main` — not for the conflicts (there are none) but so the new field and the new verdict line are **in front of you**.
2. Decide the two questions above.
3. Re-run with and without `GITHUB_TOKEN`.

Full context in #60's verdict comment; the parked documentation debt is #61.

txemi:
## Review verdict — clean; merging

Copilot's review came back empty on every push (quota), so **five independent adversarial passes** stood in for it. Their verdicts and everything applied are in this thread. The last one found nothing blocking.

### What the passes actually caught, because the list is the argument for having run them

- A **false green from a stray space**: `--own " owner "` passed the emptiness check and then matched nothing — the run went green *and* said the budget was stale.
- An **exemption marker that reached across a blank line**, suppressing a real `web_mismatch` and turning an exit 4 into a pass.
- `re.IGNORECASE` without `re.ASCII` folding U+0131, so **`gıthub.com/evil` resolved to owner `evil`**.
- The budget's outcome word **overwriting `integrity failure`**, so a failing run read as a tolerated one.
- A correction that **diverged from the merged spec in silence** and rewrote the test to assert the divergence. Amending the requirement was the resolution; FR-013 now has four branches and FR-001 names both departures from byte-identity.
- And last, the one four passes missed: a **UTF-8 BOM in the destination** hid its frontmatter, so a file that *has* a `uuid` was reported as lacking one — exit 4, over an instruction nobody can follow. This repo already carried `tests/test_bom.py` declaring that invariant for the local read paths; the web path had been left out, and 016 promoted the consequence from informational to a gate cut.

### On the tests

**306, and mutation-checked rather than counted.** That discipline earned itself here: three separate passes found a test named for a property that did not fix it — including one that passed with the exact no-op it was named after, and one that passed against the broken code because an injected fetcher returns text already decoded. Every guard in this PR now has a mutation that kills it, and the BOM fix's two halves die separately.

### Verified rather than assumed

The branch was **rebuilt on `main` twice** while under review, once reintegrating #60 by hand over the same files. Checked: `token_would_help` intact in all three sites, the anonymous-403 message identical character for character, the CHANGELOG losing nothing against the release merged in between, and — across 13,500 cases with the feature off — no change to kind, exit code, files on disk or report text. `--json` gains three keys, which FR-001 now declares.

PR #60, opened 2026-08-13
recipe: say it when the web axis cannot verify — and correct what the file claimed

txemi:
## The claim that was wrong

The header said **"public destinations need NO token"**. Public targets do not need one for **permission** — they need one for **quota**:

> Anonymous GitHub API calls are capped at **60/h per public IP**, shared by every machine behind the same NAT.

So the web axis silently stops verifying **depending on what time the build runs**, and reports `ok 0 | unverifiable N` — which reads as *"nothing to check"* and means *"could not look"*.

**Worse: it is indistinguishable from a repo that genuinely has no cross-repo links.** Nothing about the number tells you which one you are looking at.

## Measured on two different repos

| Repo shape | without token | with token |
|---|---|---|
| public, links only to itself | `ok 0` · unverifiable 52 | **`ok 7`** |
| links to a private repo | `ok 0` · unverifiable 18 | **`ok 5`** |

Both had the axis **enabled**, and neither was verifying anything.

## The change

Deliberately **not** a failure — committing offline must keep working. The gate now prints on stderr when it ends up without a token:

```
darnlink-gate: web axis running WITHOUT a token → anonymous 60/h-per-IP quota.
  'ok 0' below means COULD NOT LOOK, not 'nothing to verify'. Quote the token
  condition next to any web figure, or it is not comparable between runs.
```

**That last line is the point.** A web figure without its token condition beside it cannot be compared between two runs, let alone between two repos — and comparing them is the whole reason the axis reports a number.

Also corrected: the two header blocks that told the reader public targets are tokenless. **A comment that is wrong about the mechanism is worse than no comment**, because it is what someone reads instead of measuring.

## Verification

```
bash -n                 -> clean
without token           -> warning printed, rc=0
with token              -> no warning, ok 0 -> 5, rc=0
repo's own gates        -> 248 tests pass, lang-gate OK, dangling 0
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,27 @@ All notable changes to darnlink are documented here. The format is based on
 
 ## [Unreleased]
 
+### Fixed
+- **A tokenless `403` is the anonymous rate limit, not a private repo — and the web axis now says so
+  instead of printing a quiet `clean`.** `web-check` reported *"cannot read destination: private repo
+  and no token provided"* for any `403` without a token. That was wrong every time: GitHub answers
+  **404**, not 403, for a private repo you cannot see — as `_classify` already explained thirteen
+  lines below the message. A tokenless 403 is the **60/h-per-public-IP anonymous quota**, shared by
+  every machine behind the same NAT, so the message sent readers hunting for a permissions problem
+  they did not have, on destinations that were public and returned 200 in a browser.
+
+  **This is not only a mislabel: it can be a false green.** An un-anchored web link is only
+  discoverable if the destination can be **read**. Measured on one tree, minutes apart: *without*
+  token `exit 0`, *with* token `exit 3` — the link needed anchoring and the tokenless run could not
+  tell. So `ok 0 | unverifiable N` means *"could not look"*, not *"nothing to verify"*, and it is
+  indistinguishable from a repo with no cross-repo links at all.
+
+  Three surfaces changed accordingly: the finding text now names the quota and the remedy; the
+  summary line no longer says plain `clean` when anything was unverifiable; and both recipes (bash
+  and PowerShell) warn when the axis runs without a token. The docs that promised *"public repos work
+  tokenless"* and *"never a false green"* are corrected — **give the axis a token even for public
+  destinations**: private ones need it for permission, public ones for quota.
+
 ### Added
 - **The English-only gate now covers the surfaces that are actually published**: commit messages,
   PR titles/descriptions, issues, and `.md` documentation. Until now it judged one thing —
--- a/README.md
+++ b/README.md
@@ -25,8 +25,9 @@ that get relocated and refactored over time.
 >
 > Beyond local links, darnlink can also anchor and verify **cross-repo web links** — a Markdown link
 > to a `https://github.com/owner/repo/blob/…` file in *another* repository, anchored to that file's
-> `uuid`. The opt-in **`web-check --online`** command (off by default; tokenless for public
-> destinations) anchors a plain link and verifies an anchored one, failing on drift. It's
+> `uuid`. The opt-in **`web-check --online`** command (off by default; **give it a `GITHUB_TOKEN`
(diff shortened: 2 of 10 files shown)


txemi:
## Verdict — three adversarial rounds, one row each, and a decision to stop

Copilot's review here is **empty: quota exhausted** (`reviews=1`, body says *"unable to review … reached their quota limit"*), and it predates the last three commits. A subsession reviewer stood in for it each time, and its output was treated the same: apply what is valid, justify what is dismissed, verdict as a comment.

| Round | Findings | Applied | The one that mattered |
|---|---:|---:|---|
| **1** | 10 | 9 | The error message **contradicted the fix three lines later**: `weblinks.py` reported *"private repo and no token provided"* for a tokenless 403, while the same function explains that a private repo answers **404**. That string is where the wrong comment came from |
| **2** | 13 | 9 | `cli.py` printed **`clean`** over a pass that verified nothing. Every earlier fix touched the finding text, both recipes, the README and the doc — and left the one word that summarises the verdict |
| **3** | 9 | 3 + 1 issue | **This PR committed the defect it exists to correct** — see below |

### The round-3 finding, because it is the whole lesson

The verdict line added in round 2 printed `NOT verified (export GITHUB_TOKEN)` **unconditionally**. But `web_unverifiable` has **seven causes** and a token fixes **two**. Measured on darnlink's own tree — the configuration this PR tells everyone to adopt:

```
with token:    unverifiable 14 ... NOT verified (export GITHUB_TOKEN)
without token: unverifiable 14 ... NOT verified (export GITHUB_TOKEN)
```

All 14 are `not a recognised GitHub blob/raw URL`. **Identical figure either way**: the tool was telling the operator to export a token they had already exported, over findings it could not change.

That is the same failure the warning was added to prevent, one floor down — **an alert that fires when it cannot help is learned and ignored** — and it would have fired forever in any repo linking to non-GitHub URLs.

Round 3 also found a fix I had **declared applied and not committed** (`recipes/darnlink-gate.ps1` was absent from the commit that claimed it, while its comment asserted *"same wording … so the two surfaces cannot drift"*). Claiming an unapplied fix is worse than not applying it, because it stops anyone looking.

### Why this stops at three rounds

The rule is *"if the mechanism is still changing after three rounds, stop and split"*. Measured:

- **The code stabilised**: ~21 lines of Python across all three rounds, two files, no design change since round 2.
- **The prose did not**, and always in the same shape: a claim in a neighbouring file that the measurement contradicts — 21 lines above the fix in round 2, 29 lines above the annotation in round 3, and a sibling README that this PR itself made false.

Each fix displaces the problem one file over, and **no gate covers prose**. So the closing criterion applied is the third one — *a deterministic gate covers the change*: this PR now ships **three tests for the verdict line**, which makes a fourth round over that code redundant. The documentation debt is parked in **#61**, not repeated.

### Dismissed with reason

- **`_repo_accessible`'s 403 comment** — the same misunderstanding survives there, but it degrades *safely* (a genuine 404 becomes unverifiable). Comment corrected, behaviour deliberately untouched: changing it is a separate decision with its own blast radius.
- **The historical CHANGELOG entries** carrying the old phrasing — a changelog records what was believed at each release; rewriting it erases when the error was found.
- **The `v0.7.0` pin family** — a policy decision (the recipe's own header argues *against* printing concrete pins), not a find-and-replace. → #61.

### Validation, declared and not assumed

```
251 tests pass (3 new, covering the verdict line)
lang-gate baseline 0 · darnlink self-check: 0 repairs, 0 conflicts, dangling 0
CI 24/24 SUCCESS · mergeStateStatus CLEAN
darnlink's own tree, with and without token -> same line, no impossible advice
tokenless 403 fixture -> "1 of them would resolve with GITHUB_TOKEN"
no consumer parses `exit 0 (clean)` — swept the whole repo before changing it
```

**Checks that could not be reproduced locally:** the PowerShell recipe was reviewed by reading only — there is no `pwsh` on this machine. Its warning block is now literally identical to the bash one, but it has never been executed.

**Merging.**

PR #62, opened 2026-08-13
dangling: a link with empty text was invisible, not unreported (FR-051)

### LINKED ISSUE #52: dangling misses images with pandoc attributes: ![](x){width="1.1in"} passes the strict wall
## What happens

`darnlink check --dangling repo` **does not detect** a broken image link when it carries the
attribute suffix pandoc generates (`{width="…"}`, `{height=…}`). The target does not exist and the
gate comes out clean.

## Minimal reproduction (with the pinned binary, v0.20.4)

```
fx/README.md:   ![](media/image1.jpeg){width="1.1in"}  +  ![alt](another_missing_one.jpg)
$ uvx --from git+https://github.com/txemi/darnlink@v0.20.4 darnlink check fx
  [dangling]  targets that do not exist: 1        <-- sees ONLY the second one

fx2/README.md:  only ![](media/image1.jpeg){width="1.1in"}
$ … darnlink check fx2
  -> exit 0 (clean)                               <-- the [dangling] line is not even printed
```

Neither target exists. The left-hand one is invisible.

## Why this matters more than it looks

In a real repo running the wall at maximum (`mode: max`, `dangling: repo`, no `dangling_max`), a
sweep found **7 links with a `{attr}` suffix and a non-existent target, and 0 with `{attr}` and a
valid target**. In other words: in that corpus **every** link of that shape is broken, and **every**
one of them is invisible to the gate.

The pattern comes from converting `.docx`/`.odt` to Markdown with pandoc, so it shows up in whole
blocks of files imported at once — this is not an isolated case.

The effect is the worst kind a gate can have: **a `dangling: 0` that does not mean there are no
dangling links**, only that there are none *of the shapes the regex recognises*. Whoever reads it
concludes the former.

## Where to look

The image regex: the part that accepts `alt` appears not to allow for a `{…}` suffix
immediately after it, so the whole link is discarded instead of keeping the `dest`.

## Suggestion

Accept (and ignore) an optional `{…}` attached to the closing parenthesis, and add a test with both
variants of the fixture above — the one with a valid target and the one without — so a case where
the link is discarded instead of checked cannot slip through again.

txemi:
Closes #52.

## What was broken

`MD_LINK_RE` required at least one character of link text (`[^\]]+`), so a link whose text is empty
— ``, `[](doc.md)` — **never matched**. Not a finding that got filtered: never a
candidate, so no axis ever saw it.

The axis then printed `dangling: 0` over a tree containing broken image embeds. That reads as *"no
broken links"*; it only ever meant *"none of the shapes the regex recognises"*, and nothing in the
output distinguishes the two. The empty alt is what document converters emit for images, so it
arrives in whole blocks of files imported at once — the kind nobody re-reads.

**The reported cause was not the cause.** #52 blamed the pandoc attribute suffix (`{width="1.1in"}`);
measured against the regex, the suffix was always harmless and the empty text was the whole of it.
Both are pinned in tests so the true cause cannot be re-diagnosed from the same coincidence.

## Measured

One repository at `mode: max`, **in its own gate scope** (excludes + ignore-blocks):

| | |
|---|---|
| empty-text links in scope | **127** |
| → non-local (http/mailto/fragment) | 37 |
| → local target **exists** | 83 |
| → local target missing | **7** — every one a real broken image embed |
| `dangling` before → after | **0 → 7** · findings lost: **0** |

## The change

- `[^\]]+` → `[^\]]*` in `MD_LINK_RE` **and** `ROBUST_LINK_RE`. They widen together: leaving the
  robust one narrow would make an anchored `[](path) <!-- uuid: … -->` plain to one finder and robust
  to the other, and that coupling is load-bearing for the **repair** axis.
- **FR-052**, within the `dangling` axis: the whitespace *delimiting* a destination is not part of
  it. Four properties of that rule each took a review round to get right, and each is a false green
  when wrong — the full statement is in the CHANGELOG. In short: strip **before** decoding (an escape
  is content, `[x](%20a.md)` denotes `" a.md"`), **ASCII only** (NBSP is an ordinary character), the
  **two edges differ** (`cmark` keeps VT/FF at the end), and the rule is **POSIX-shaped**.

## ⚠️ Adopting this is not a no-op

`MD_LINK_RE` has three call sites. In the same repository:

- **web axis: +36 links newly visible.** None is a `/blob/` URL today, so its web gate does not flip
  — the shape of those URLs, not a property of the change. A `/blob/` URL exits **3** if the
  destination carries a uuid and **4** if it 404s with a token that can see the repo.
- **`--create-readme`**: `[](sub/)` and `` can now create a `README.md`. 0 occurrences.
- **the operational one**: a consumer at `dangling: repo` with `dangling_max` unset goes **0 → 7 and
  its push wall closes**. Fix the links or raise the ceiling *before* moving the pin.

## Known issues, filed rather than folded in

Four, all pre-existing, listed in the CHANGELOG where an upgrader will see them: **#71** — *live*,
104 links reported dead while the file exists, because balanced parentheses in a destination are
truncated · **#65** `--write` detaches a pandoc attribute block · **#67** whitespace is stripped by
`dangling` but not by `repair`/`robustify`/`--create-readme` · **#68** `--write` drops the UTF-8 BOM.

## Verification

- **274 tests pass** (was 251 on `main`).
- **Eight adversarial review rounds**, none of which was allowed to report clean without first
  seeding a defect and proving the suite catches it. Ledgers are comments on this PR.
- Across thirteen repositories and several differential fuzzes: **0 findings lost, 0 writes changed**.
  The only deliberate subtraction is FR-052 — 0 whitespace-only destinations and 6 with surrounding
  whitespace, none of which changes a verdict.
- Round 7 and round 8 adjudicated the whitespace rule against the **reference CommonMark
  implementation** rather than the spec prose, and each overturned a claim the previous rounds had
  propagated. The lesson is written into the spec: when a rule is about syntax, ask the parser.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,8 +33,9 @@ All notable changes to darnlink are documented here. The format is based on
   so an anchored empty-text link cannot be plain to one function and robust to another — a coupling
   that matters to the **repair** axis and has its own tests, since reverting that half alone leaves
   every `dangling` test green. FR-051.
-- **The whitespace around a link destination is no longer treated as part of it**, as CommonMark
-  requires. Two halves:
+- **Within the `dangling` axis**, the whitespace around a link destination is no longer treated as
+  part of it, as CommonMark requires. (Only that axis — `repair`, `robustify` and `--create-readme`
+  still do not strip; see *Known issues* and #67.) Two halves:
 
   *A destination that is only whitespace is not reported at all.* `[x]( )`, `[x](%20)` and
   `[x]( #sec)` resolved to the linking file's own directory under a blank name, giving a finding
@@ -46,7 +47,11 @@ All notable changes to darnlink are documented here. The format is based on
   denotes `B.md`; judged literally it resolved to `dir/ B.md ` and was reported dead while the file
   sat right there — a **false red on a live file**, which is how a gate gets switched off rather
   than fixed. Applied to the decoded path with the fragment removed, and — the part a first version
-  got wrong — at every place that value is *used*, not only where it is tested. FR-052.
+  got wrong — at every place inside this axis that the value is *used*, not only where it is tested.
+  **ASCII whitespace only**, which is what CommonMark counts: a bare `str.strip()` also removes NBSP
+  and the ideographic space, and since a file really can be named `\xa0a.md`, that turned a link to
+  a **missing** file into a resolved one and reported clean — a false green, and NBSP is precisely
+  what a Word or HTML paste produces. FR-052.
 
   Note this **removes** findings from the pre-existing surface rather than adding any: with a
   non-empty text these shapes were reported before this release. Measured across thirteen local
@@ -75,18 +80,23 @@ All notable changes to darnlink are documented here. The format is based on
 
 ### Known issues — read this before moving a pin
 
-Neither is introduced by this release and both are latent (**0 occurrences across thirteen
+None of the three is introduced by this release and all are latent (**0 occurrences across thirteen
 repositories**), but this release routes more links towards the first, so they belong where a
 consumer deciding on an upgrade will see them — not buried under *Fixed*.
 
 - **`--write` detaches a pandoc attribute block** (#65). The anchor lands between the link and its
   `{…}`, so the block stops applying. Pre-existing: a non-empty link text has always done it. The
   tests added here pin that the block is never *deleted*, which is the worse neighbour of the two.
+- **`--write` silently drops a file's UTF-8 BOM** (#68) on all four write paths. CRLF is preserved
+  meticulously; the BOM is not, and three places in this repo imply otherwise — including the CI
+  Windows matrix, whose stated purpose is BOM/CRLF and whose BOM fixture only covers the *read*
+  path, so it cannot see this.
 - **Whitespace around a destination is stripped by `dangling` but not by `repair`, `robustify` or
   `--create-readme`** (#67), so one link can get two verdicts. The `repair` case is the sharp one:
   a trailing space makes `names_md` false, the link is classified as a directory link and becomes a
   `CONFLICT` diagnosed as *"path and uuid disagree"* — which is untrue, and `--write` never heals
   it, so the gate stays red. FR-052 is deliberately scoped to the `dangling` axis for now.
+
 ### Added
 - **The English-only gate now covers the surfaces that are actually published**: commit messages,
   PR titles/descriptions, issues, and `.md` documentation. Until now it judged one thing —
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,12 +46,19 @@ All notable changes to darnlink are documented here. The format is based on
   *A destination with whitespace **around a real name** is stripped before resolving.* `[x]( B.md )`
   denotes `B.md`; judged literally it resolved to `dir/ B.md ` and was reported dead while the file
   sat right there — a **false red on a live file**, which is how a gate gets switched off rather
-  than fixed. Applied to the decoded path with the fragment removed, and — the part a first version
-  got wrong — at every place inside this axis that the value is *used*, not only where it is tested.
-  **ASCII whitespace only**, which is what CommonMark counts: a bare `str.strip()` also removes NBSP
-  and the ideographic space, and since a file really can be named `\xa0a.md`, that turned a link to
-  a **missing** file into a resolved one and reported clean — a false green, and NBSP is precisely
-  what a Word or HTML paste produces. FR-052.
+  than fixed. FR-052.
+
+  Two properties of that strip took four attempts to get right, and both are false greens when
+  wrong, so they are worth stating exactly:
+
+  - **It happens BEFORE percent-decoding.** The whitespace dropped is a *delimiter*, and a delimiter
+    exists only in the source. An escape is content: `[x]( a.md )` denotes `a.md`, but `[x](%20a.md)`
+    denotes `" a.md"` — checked against the reference CommonMark implementation, which emits
+    `href="a.md"` for the first and `href="%20a.md"` for the second. Stripping afterwards made them
+    the same link, so a link to a **missing** file was answered by an existing neighbour.
+  - **ASCII whitespace only**, which is what CommonMark counts. A bare `str.strip()` also removes
+    NBSP and the ideographic space; a file can be named `\xa0a.md`, so that hid a broken link too —
+    and NBSP is precisely what a Word or HTML paste produces.
 
   Note this **removes** findings from the pre-existing surface rather than adding any: with a
   non-empty text these shapes were reported before this release. Measured across thirteen local
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,9 +33,16 @@ All notable changes to darnlink are documented here. The format is based on
   so an anchored empty-text link cannot be plain to one function and robust to another — a coupling
   that matters to the **repair** axis and has its own tests, since reverting that half alone leaves
   every `dangling` test green. FR-051.
-- **`[]( )` — an href of pure whitespace — was reported as a dangling target.** It resolved to the
-  linking file's own directory with a blank name, so the finding read `line 5:  : target does not
-  exist`, naming nothing anyone could go and look for. FR-052.
+- **An href that names no destination is no longer reported.** `[x]( )`, `[x](%20)` and `[x]( #sec)`
+  all resolved to the linking file's own directory under a blank name, giving a finding that read
+  `line 5:  : target does not exist` — naming nothing anyone could go and look for. The last one was
+  worse than cosmetic: ` #section` is a legal in-page anchor, and FR-046 says a bare fragment is
+  never reported, so a valid link was being called dead. Judged on the **decoded** path after the
+  fragment is split off, because the three spellings are one link. FR-052.
+
+  Note this **removes** findings from the pre-existing surface rather than adding any: with a
+  non-empty text these shapes were reported before this release. Measured across eleven local
+  repositories: **0 occurrences**, so no consumer's count moves.
 - **A tokenless `403` is the anonymous rate limit, not a private repo — and the web axis now says so
   instead of printing a quiet `clean`.** `web-check` reported *"cannot read destination: private repo
   and no token provided"* for any `403` without a token. That was wrong every time: GitHub answers
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,6 +56,13 @@ All notable changes to darnlink are documented here. The format is based on
     denotes `" a.md"` — checked against the reference CommonMark implementation, which emits
     `href="a.md"` for the first and `href="%20a.md"` for the second. Stripping afterwards made them
     the same link, so a link to a **missing** file was answered by an existing neighbour.
+  - **The two edges are not the same set.** `cmark` strips VT and FF at the start of a destination
+    and keeps them at the end (`[x](a.md\x0b)` → `href="a.md%0B"`), so a symmetric strip is itself a
+    false green. Leading: space, tab, LF, CR, VT, FF. Trailing: space, tab, LF, CR.
+  - **The rule is POSIX-shaped.** Windows trims trailing spaces from a path component, so
+    `[x](a.md%20)` normalises to `a.md` there and is not dangling. That is the filesystem's answer
+    and this axis exists to ask the filesystem — but a consumer should read it here, not discover it
+    from a build that is red on one OS only.
   - **ASCII whitespace only**, which is what CommonMark counts. A bare `str.strip()` also removes
     NBSP and the ideographic space; a file can be named `\xa0a.md`, so that hid a broken link too —
     and NBSP is precisely what a Word or HTML paste produces.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ All notable changes to darnlink are documented here. The format is based on
   `dangling: 0`. Since those files are converted documents, they arrive in blocks and nobody
   re-reads them.
 
-  ⚠️ **Adopting this is not a no-op for a consumer.** `MD_LINK_RE` is shared by four call sites, so
+  ⚠️ **Adopting this is not a no-op for a consumer.** `MD_LINK_RE` has three call sites, so
   the same repository also gains **36 newly visible web links** (none a `/blob/` URL today, so its
   web gate does not flip — the shape of those URLs, not a guarantee), and `--create-readme` gains a
   path where `![](media/)` can create a `README.md`. A repo at `dangling: repo` with `dangling_max`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -94,9 +94,18 @@ All notable changes to darnlink are documented here. The format is based on
 
 ### Known issues — read this before moving a pin
 
-None of the three is introduced by this release and all are latent (**0 occurrences across thirteen
-repositories**), but this release routes more links towards the first, so they belong where a
-consumer deciding on an upgrade will see them — not buried under *Fixed*.
+None is introduced by this release, but they belong where a consumer deciding on an upgrade will
+see them rather than buried under *Fixed*. **One of them is live**; the rest are latent at 0
+occurrences across thirteen repositories.
+
+- ⚠️ **LIVE — balanced parentheses in a destination are truncated at the first `)`** (#71).
+  CommonMark allows them; `MD_LINK_RE` does not, so the link is cut short and reported dead while
+  the file it names sits on disk. **104 occurrences in one repository**, clustered in mirrored
+  attachment filenames (`(Parte 1)`, `(February - Monthly)`). The report conceals itself: its own
+  `(resolves to …)` supplies the missing parenthesis, so the truncated path reads as complete, and
+  a reader who checks finds the file present and concludes the *gate* is broken. That repo runs
+  `dangling: warn`, so these print rather than gate — **raising it to `repo` would close the wall on
+  104 files that exist.** Pre-existing and orthogonal to this release.
 
 - **`--write` detaches a pandoc attribute block** (#65). The anchor lands between the link and its
   `{…}`, so the block stops applying. Pre-existing: a non-empty link text has always done it. The
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,22 @@ All notable changes to darnlink are documented here. The format is based on
 ## [Unreleased]
 
 ### Fixed
+- **A link with empty text — `![](photo.jpg)`, what pandoc emits for every image in a converted
+  `.docx`/`.odt` — was invisible to every axis, not merely unreported.** `MD_LINK_RE` required at
+  least one character of link text (`[^\]]+`), so such a link never matched: it was not a finding
+  that got filtered, it was never a candidate. The axis then printed `dangling: 0` over a tree full
+  of broken image embeds, which reads as *"no broken links"* but only ever meant *"none of the
+  shapes the regex recognises"*.
+
+  Measured on one repository running the wall at maximum: **7 links of this shape, 7 of them broken,
+  0 visible** — and none with a valid target, so the corpus offered no counter-example either. Since
+  those files are converted documents, they arrive in blocks and nobody re-reads them.
+
+  Reported as the pandoc attribute suffix (`{width="1.1in"}`) hiding the link; it was not.
+  The pattern stops at the `)` and never looks past it, so `![alt](x.jpg){width="1.1in"}` was always
+  seen. The two shapes simply co-occur. Both are pinned in tests so the real cause — the empty text —
+  cannot be re-diagnosed from the same coincidence. `ROBUST_LINK_RE` widens alongside `MD_LINK_RE`,
+  so an anchored empty-text link cannot be plain to one function and robust to another. FR-051.
 - **A tokenless `403` is the anonymous rate limit, not a private repo — and the web axis now says so
   instead of printing a quiet `clean`.** `web-check` reported *"cannot read destination: private repo
   and no token provided"* for any `403` without a token. That was wrong every time: GitHub answers
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -100,12 +100,25 @@ occurrences across thirteen repositories.
 
 - ⚠️ **LIVE — balanced parentheses in a destination are truncated at the first `)`** (#71).
   CommonMark allows them; `MD_LINK_RE` does not, so the link is cut short and reported dead while
-  the file it names sits on disk. **104 occurrences in one repository**, clustered in mirrored
-  attachment filenames (`(Parte 1)`, `(February - Monthly)`). The report conceals itself: its own
-  `(resolves to …)` supplies the missing parenthesis, so the truncated path reads as complete, and
-  a reader who checks finds the file present and concludes the *gate* is broken. That repo runs
-  `dangling: warn`, so these print rather than gate — **raising it to `repo` would close the wall on
-  104 files that exist.** Pre-existing and orthogonal to this release.
+  the file it names sits on disk, clustered in mirrored attachment filenames (`(Parte 1)`,
+  `(February - Monthly)`). The report conceals itself: its own `(resolves to …)` supplies the
+  missing parenthesis, so the truncated path reads as complete, and a reader who checks finds the
+  file present and concludes the *gate* is broken.
+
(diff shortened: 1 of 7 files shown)


txemi:
## Review ledger — round 1

Copilot returned *"reached their quota limit"*, which counts as `reviews=1` and shows nothing. Per
REGLA CERO the substitute is a Claude subsession review, treated exactly like Copilot's: apply what
is valid, justify what is dismissed, verdict written here.

**Round 1 could not break the core change.** Over four real repositories and a 200k-string
differential fuzz through the real `find_plain_links` / `find_robust_links` / `find_detached_anchors`
APIs with code spans applied: **0 findings lost, 0 writes changed**. What it broke was the *evidence
around* the change — which is worth more, because a gate is only as good as the claims made for it.

| # | Finding | State | What was done |
|---|---|---|---|
| 1 | **Half the diff had no test.** Reverting only the `ROBUST_LINK_RE` widening left the suite green — all four new tests reach `MD_LINK_RE` via `find_plain_links` | **resolved** | 3 tests in `test_links.py`. Seeded the revert: **3 failures where there had been 0** |
| 2 | **The headline measurement was false** — and falsely reassuring | **resolved** | Re-measured in the repo's own gate scope. See below |
| 3 | **Blast radius undeclared**; adoption is not a no-op for consumers | **resolved** | Table in the spec + warning in CHANGELOG and PR body |
| 4 | `[]( )` produced a finding reading `line 5:  : target does not exist` | **resolved** | FR-052 guard; seeded its removal → 1 failure |
| 5 | Stale claim in FR-041: *"no finding kind has ever carried a line field"* | **resolved** | `Finding` grew `line` (`report.py:37`), `robustify.py:397` fills it. Corrected in place |
| 6 | Test docstring claimed to pin the non-empty-alt case; no test used one | **resolved** | Added that test |
| 7 | Lost-match shape `[]([a](b.md))` — outer match swallows the inner link | **dismissed with reason** | Exhaustive search over the alphabet bounds *every* lost-match string to the literal prefix `[](` + a nested link. No realistic Markdown does this, and guarding it would cost more than it buys |
| 8 | New false-positive *instances* of pre-existing classes: `\[\](x)`, links inside HTML comments, `.png)` truncation | **dismissed with reason** | These classes predate this PR and fire identically with non-empty text. Fixing them is a different feature (the `href` grammar), and folding it in here would hide a behaviour change inside a regression fix |
| 9 | Detached-anchor absorption changes in one contrived shape | **dismissed with reason** | Semantically correct under the widened grammar, and `DetachedAnchor` is not a reported `Kind`, so no finding disappears |

### On finding 2, because it is the one worth writing down

The PR claimed **`132 → 139`** and *"0 links of this shape with a target that exists, so the corpus
offered no counter-example"*. Re-measured, independently, twice:

| | claimed | actual |
|---|---:|---:|
| empty-text links in gate scope | 7 | **127** |
| → with a target that **exists** | **0** | **83** |
| → broken | 7 | 7 ✓ |
| `dangling` before → after | 132 → 139 | **0 → 7** |

Two separate errors, and they fail in the same direction:

- **`132`** came from a run that omitted the repo's `--ignore-block` flags, so it counted 132
  findings the real gate deliberately ignores. The gate's own scope is the only scope that means
  anything, and it was not the one measured.
- **"no counter-example"** came from having measured a *narrower* shape than the sentence named —
  images carrying a pandoc suffix, not empty-text links. Both were called *"this shape"*, which is
  what made the gap invisible.

A widening is justified by **what it newly matches**. Counting the subset that motivated it and
presenting that as the total does not merely understate the change: it inverts the safety argument,
turning "83 links I did not look at" into "nothing else exists". That is the same species of error
as the bug being fixed — a number that is read as *"nothing there"* when it means *"nothing counted"*.

### Not reproducible in local CI

None. This repo is public and its Actions run: 21 checks green, including the `scan` privacy-gate and
`lang`. Nothing here depends on a CI secret.

**Round 2 is running** on the fixes above, since applying findings produces code nobody has reviewed.

txemi:
## Review ledger — round 2

Round 2 re-derived **every** number from round 1's corrections and found them exact: 127 / 37 / 83 /
7 empty-text links, 36 web, 0 `--create-readme` occurrences, `0 → 7` dangling. Both confessions in
the spec were confirmed accurate, including that the retracted `132 → 139` reproduces exactly when
the ignore-blocks are omitted.

**What it broke was the tests round 1 added.** Which is the more valuable outcome: those tests were
the evidence that round 1's fix worked.

| # | Finding | State | What was done |
|---|---|---|---|
| 1 | **The suffix test caught nothing.** Seeded the wrong fix it claims to guard — a regex ending `\)(?:\{[^}]*\})?`, which *consumes* the suffix — and **all 260 tests passed** | **resolved** | The property is only observable on the **write** path: with that seed, `--write` silently **deletes the attributes** from the document, while detection reports an identical finding. Test moved to `test_robustify.py`; the same seed now fails |
| 2 | **The `!`-preservation invariant was asserted in three documents and tested by none** — the highest-value gap, since this widening is exactly what routes empty-alt embeds into `--write` for the first time | **resolved** | Write test added |
| 3 | **FR-052's guard read the RAW href**, five lines above a comment stating the opposite rule. `[](%20)` and `[]( #sec)` still emitted the forbidden finding — and the second is a **false positive on valid CommonMark**, since FR-046 says a bare fragment is never reported | **resolved** | Guard moved below `split_fragment` + `unquote`; two tests, both seeded |
| 4 | `test_emit_round_trips_an_empty_text_link` had a **dead conditional** and a docstring describing span arithmetic the test never ran | **resolved** | Scope narrowed to what it does, pointing at the test that does the rest |
| 5 | Three docstrings asserted *"every other test in this file uses an empty alt"*. The file has 26 tests; **at least 12 use a non-empty text** | **resolved** | Corrected |
| 6 | One copy of the **retracted** measurement survived in a test docstring | **resolved** | Corrected. It was in the one surface a maintainer greps first |
| 7 | `--write` **detaches a pandoc attribute block**: the anchor lands between the link and its `{…}` | **declared, filed as #65** | Measured on both shapes: a non-empty text does the same and always has, so it is **pre-existing**. What is new is the exposure. Folding a behaviour change into a regression fix is how a fix stops being reviewable |
| 8 | The `/blob/` exit-code claim was incomplete | **resolved** | Exits **3** if the destination reads and carries no uuid; **4** — integrity failure, the harder stop — if it 404s with a token |
| 9 | Spec rows sat under *"Measured on the same repository"* carrying no measurement | **resolved** | The `robustify --write` row is **0** in that corpus: all 83 empty-text links with a live target point at `.jpg`/`.jpeg`, which are unanchorable |
| 10 | Unverifiable fleet-wide figures elsewhere in the spec (3.212 dangling, "six of fourteen") | **dismissed with reason** | They predate this feature and this PR does not touch them. Correcting numbers this PR did not author would hide the ones it did |

### The lesson worth keeping, because it is the same one twice

Round 1's finding was a **measurement** that described a narrower thing than its sentence claimed.
Round 2's headline finding is a **test** that described a stronger property than its assertion
checked. Same shape, one level down: in both cases the artifact looked like evidence and was not,
and in both cases nothing in the output distinguished the two.

The concrete rule that falls out: **for anything on a write path, a test that only inspects findings
proves nothing.** Detection is identical whether or not the suffix is inside the match span — the
documents differ, and only a write shows it. That is precisely how this survived round 1.

### Not reproducible in local CI

None. Public repo, Actions run, 25 checks green including the `scan` privacy-gate.

**Round 3 is running**, on round 2's fixes, with an explicit instruction to seed a defect and prove
the suite catches it before reporting anything as clean — the distinction between *dry* and *blind*.

txemi:
## Review ledger — round 3, and the verdict

Round 3 confirmed the FR-052 guard move is sound (28 edge hrefs probed — `.`, `./`, `%2520`, `%09`,
`\xa0`, `%E2%80%83`, … — **no input where it suppresses a link that could name a real file**), and
independently confirmed every zero-occurrence claim over **13 repositories, 14.446 Markdown files and
52.932 in-prose links**. Then it broke round 2's own work.

| # | Finding | State | What was done |
|---|---|---|---|
| 1 | ⚠️ **The suffix test was tuned to its own seed.** A wrong fix one character narrower — `\)(?:\{\.[^}]*\})?`, consuming only **class** blocks — passes **264/264** while `--write` deletes `{.cls}` from the document | **resolved** | Fixture now covers four attribute shapes. The class seed **and** an anchor seed nobody had tried both fail against it |
| 2 | The `!` test's only load-bearing assertion was `a.startswith("!")`, true only while the fixture began with the link | **resolved** | Asserts the bang is attached to *this* link, with a fixture that no longer starts with it. (`"!" in a` would be worse — the emitted `<!--` contains one) |
| 3 | ⚠️ **The exit-code table was INVERTED.** `web_anchor` — exit 3 — fires when the destination **carries** a uuid, not when it lacks one; and 404-with-token exits 4 only if the token can see the repo | **resolved** | Replaced with the four cases read off the code |
| 4 | Inserting the #65 paragraph mid-table **terminated the table**: the last row rendered as literal pipe-text | **resolved** | Paragraph moved out. Verified by rendering: 2 tables, 11 `<tr>`, 9 data rows |
| 5 | **FR-052 was incomplete by its own stated criterion**: `[x]( B.md )` denotes `B.md` under CommonMark but resolved to `dir/ B.md ` and was reported dead | **resolved** | Strips rather than merely testing for empty. Its normative sentence also called the three spellings "one link", false under CommonMark — they warrant the same *treatment*, which is what it now says |
| 6 | *"`MD_LINK_RE` is shared by four call sites"* — it has three | **resolved** | Corrected in CHANGELOG and spec |
| 7 | A cross-reference to "the write case below" that round 2 had moved to another file | **resolved** | Named in full |
| 8 | The PR body still carried round-1 wording on three points | **resolved** | Rewritten |

### The seeding, because a round that cannot catch anything is not a clean round

| Seed | Caught? |
|---|---|
| regex consuming a **class** block `{.cls}` | ✅ 1 failure *(passed 264/264 before this round)* |
| regex consuming an **anchor** block `{#id}` — untried by anyone until now | ✅ 1 failure |
| span that eats the `!` of an embed | ✅ 1 failure |
| removing the new `strip()` from FR-052 | ✅ 1 failure |

**266 tests pass** (was 264).

## Verdict: closing the loop at round 3

Three rounds, none of them dry. But look at *what* stopped changing and when:

| | R1 | R2 | R3 |
|---|---|---|---|
| the fix itself (`+` → `*`) | settled | untouched | untouched |
| what the fix *claims* | false measurement | — | — |
| the **tests** | absent | wrong property | right property, one sample |
| the **documents** | undeclared radius | mis-scoped rule | inverted exit code, broken table |

**The mechanism has been stable since round 1.** Every subsequent finding was about *evidence* —
measurements, tests, prose — which is why three rounds is a reasonable place to stop rather than a
signal to split the PR: there is no unverified mechanism left to carve out.

The loop closes here on the second criterion — **no regression risk left**: what remains after this
round is documentation, and the code is now guarded by four seeded defects it demonstrably catches,
including two variants nobody had thought of before this round.

The honest caveat: closing at three rounds is a **decision**, not a proof. Round 3 was not dry, so
round 4 would probably find something too. What it would find, on this evidence, is a fifth way of
writing down something already true — not a fifth defect in the gate.

### Not reproducible in local CI

None. Public repo, Actions run, all checks green including the `scan` privacy-gate and `lang`.

### Left open, deliberately

**#65** — `--write` detaches a pandoc attribute block. Pre-existing (a non-empty text does the same
and always has), reproduced and filed rather than folded in. Frequency today: 0, because every
empty-text link with a live target points at an unanchorable image.

### ⚠️ Landing order — this is not a no-op for consumers

A repo at `dangling: repo` with `dangling_max` unset goes **0 → 7** and its push wall closes.
**Fix the links or raise the ceiling before moving the pin**, not after.

txemi:
## Review ledger — round 4

Round 4 verified the exit-code table row-by-row with an injected fetcher (round 3's inversion is
genuinely fixed, not differently wrong), rendered the spec whole (3 tables, 16 rows, nothing escaped
into a paragraph), and diffed old-vs-new findings over 13 repos and 43.187 links: **0 lost**.

Then it planted nine seeds. **Two passed 266/266 while `--write` silently deleted document content.**

| # | Finding | State | What was done |
|---|---|---|---|
| 1 | ⚠️ **The four-shape suffix test held the LINK TEXT constant** — every fixture used an empty alt. A wrong fix that eats the block only when the text is *non-empty* passes the whole suite while `alt text{width=…}` loses its attributes | **resolved** | 12 cases: 3 texts × 4 shapes |
| 2 | ⚠️ **`ROBUST_LINK_RE` has its own write path — `repair` — and nothing guarded it.** Seeding the block-eater there passes the suite while `repair --write` moves the link and deletes the block on the way | **resolved** | Two tests, one per position the block can occupy relative to the anchor. The first version covered only the position the seed cannot reach |
| 3 | ⚠️ **FR-052 was applied where it was TESTED, not where it was USED.** `stripped` was computed for the emptiness check while resolution used `href.strip()` — a different string that cannot reach a space before a `#` — and the encoded branch was never stripped at all | **resolved** | `[x]( B.md #s )`, `[x](%20B.md )` and `[x](B.md #s)` were reported dead with the file sitting right there. A **false red on a live file** is how a gate gets switched off rather than fixed |
| 4 | The `!` assertion was satisfiable while wrong: `"" in a` is true of `!` | **resolved** | Asserts the exact line; the fixture is deterministic apart from the uuid |
| 5 | **The CHANGELOG had not been carried through round 3** — four mismatches in one paragraph, including the "one link" phrasing round 3 retracted and an omission of #65, the only blast-radius item that corrupts documents | **resolved** | Rewritten |
| 6 | *"0 occurrences"* sat under a rule that fires **6** times | **resolved** | It changes no verdict, only the path a finding names — but the bare zero read as *inert*. Both numbers stated |
| 7 | A stale `robustify.py:397` reference, moved twice since it was written | **resolved** | Replaced with the function name, which does not drift |
| 8 | `_anchor_target` / `_dir_link_missing_readme` do not strip, so `[x]( B.md )` is neither dangling nor robustifiable | **carried to round 5** | 0 occurrences. It is an inconsistency, not a wrong verdict, and it is being re-attacked rather than waved through |

### The seeding, and why it is now the interesting part

| Seed | Caught? |
|---|---|
| class-block eater · anchor-block eater · mid-line-only eater | ✅ |
| revert the core widening · drop `.strip()` · `lstrip` instead of `strip` · guard on raw href | ✅ |
| **block eater, non-empty link text only** | ❌ → now ✅ |
| **`ROBUST_LINK_RE` eats the block (`repair --write`)** | ❌ → now ✅ |

**269 tests pass** (was 266). All nine seeds now fail the suite.

### The pattern, stated because it is the actual finding of this loop

Each round has beaten the previous round's tests **by varying an axis nobody varied**:

| | beat the round before by varying… |
|---|---|
| R2 | the *wrong fix* — the tests never tried one |
| R3 | the *attribute shape* — `{.cls}` where the fixture had `{width=…}` |
| R4 | the *link text* — non-empty where every fixture was empty. And by finding a **second write path** with no guard at all |

Round 4's axis is the one this whole feature is about, held constant in the tests written to guard
it. So round 5's first instruction is explicit: **find the axis round 4 held constant.**

Not reproducible in local CI: none. Public repo, Actions green including the `scan` privacy-gate.

txemi:
## Review ledger — round 5

**No false green.** A differential of `main`'s `_dangling_target` against HEAD's, over every local
repository (9.352 files, 22.149 plain links, restricted to links that actually reach the function):

```
findings LOST   (old dangling, new silent): 0
findings GAINED (old silent, new dangling): 0
path CHANGED    (same verdict, better path): 6
```

Exactly what the CHANGELOG claims. Six other things, four applied and two filed.

| # | Finding | State | What was done |
|---|---|---|---|
| 1 | ⚠️ **The repair test asserted `'width="1.1in"' in a`** — satisfied by `-->width="1.1in"}`, the brace eaten and the block silently demoted to prose. A splice one character off produces exactly that and passed **269/269** | **resolved** | Asserts the whole line, like its sibling. This is the weakness round 4 fixed in the bang test **in the same commit** — *"there is no reason to assert anything less than the output"* — and I then reintroduced it one test over |
| 2 | **The same test's docstring named a seed it does not catch** (that seed fails the *sibling*) | **resolved** | Rewritten to say what it actually guards |
| 3 | ⚠️ **The only path in `robustify` that DELETES text had no guard.** Attribute-block tests exist and detached-anchor tests exist; **no fixture combined them**, so a boundary walking back over `}` as well as whitespace eats the closing brace and passes the whole suite | **resolved** | Test added; both seeds now fail |
| 4 | The CHANGELOG filed a **document-corrupting defect under `### Fixed`** | **resolved** | Moved to `### Known issues`, where an upgrader looks. (The first attempt swallowed an unrelated `Fixed` bullet; caught and re-cut) |
| 5 | The PR body was two commits stale in three places | **resolved** | Rewritten |
| 6 | **FR-052 was written as an unqualified rule about link destinations while the code implements it in one function** | **filed as #67** | See below |

### On #6, because it is the one with a real consequence

Three other sites resolve the same href without stripping. On the `repair` axis that is not an
inconsistency, it is a **wrong verdict the tool cannot heal**:

```
href='old/B.md'   -> repair: old/B.md -> new/B.md                        rewritten
href='old/B.md '  -> CONFLICT: "path and uuid disagree; left untouched"  NOT rewritten
```

`names_md("old/B.md ")` is false — trailing space — so the link is classified as a **directory**
link, its uuid turns up in a non-README, and the result is a `CONFLICT`, which feeds `integrity_fail`
→ **exit 2**. `--write` never heals it, so the gate stays red, and the diagnosis printed is **false**:
the path and the uuid agree perfectly.

**0 occurrences across thirteen repositories** (0 of 10.116 robust links). Fixing it changes
behaviour on three axes this feature does not own, so FR-052's sentence is narrowed to *"within the
`dangling` axis"* and the gap is filed. **A gap on the record is a decision; a gap left implied is a
silence** — and this loop has spent five rounds on exactly that distinction.

### Seeding

| Seed | Caught? |
|---|---|
| `_hspace_start` walks over `}` (deletes a brace) | ❌ → now ✅ |
| repair splice off-by-one (deletes a brace) | ❌ → now ✅ |
| the twelve seeds of rounds 1-4 | ✅ |
| `names_md` strips · `_anchor_target` strips | ❌ — nothing pins the current behaviour, which is *why* #67 says to do it deliberately rather than opportunistically |

**270 tests pass** (was 269).

### A methodological note worth keeping

A same-size mutation (`[^\]]*` → `[^\]]+`) restored within the same filesystem second leaves a
`__pycache__` `.pyc` that Python still considers valid — it keys on mtime-**seconds** plus size. The
seeded code keeps running against a `git status`-clean tree and produces phantom failures. **Clear
`__pycache__` between seed and restore.** Five rounds of seeding and this surfaced only now.

Not reproducible in local CI: none. Public repo, Actions green including the `scan` privacy-gate.

txemi:
## Review ledger — rounds 6, 7 and 8

Three rounds in one comment, because they turned out to be one finding examined from three sides.
(They should have been posted as they happened; that they were not is itself a lapse, and round 8
caught it.)

### The single question these three rounds were about

**Which spellings of whitespace does FR-052 strip?** Rounds 3, 4, 6, 7 and 8 each closed one
dimension of that question and re-derived the next from the rule sitting beside it:

| Round | Dimension closed | How the next one was got wrong |
|---|---|---|
| 3 | literal whitespace | strip placed *after* decoding |
| 4 | applied where *used*, not only where tested | Unicode vs ASCII unexamined |
| 6 | **ASCII only** — `str.strip()` also eats NBSP | encoded vs literal unexamined |
| 7 | **before decoding** — an escape is content | the two edges assumed symmetric |
| 8 | **the edges are asymmetric** — `cmark` keeps VT/FF at the end | platform stated, not fixed |

**What broke the cycle was going to the reference implementation.** Rounds 3-6 reasoned from the
spec and from the adjacent requirement; rounds 7 and 8 asked `cmark`, and each immediately overturned
a claim the previous rounds had propagated:

```
[x]( a.md )    ->  href "a.md"       delimiters, dropped
[x](%20a.md)   ->  href "%20a.md"    escape: CONTENT — denotes " a.md"
[x](\x0ba.md)  ->  href "a.md"       VT leading:  delimiter
[x](a.md\x0b)  ->  href "a.md%0B"    VT trailing: CONTENT
```

Each mismatch was a **false green**: a link to a missing file answered by an existing neighbour.
The clause that licensed the first one — *"those spellings denote the same destination"* — was
borrowed from FR-047, where it is true, and reached the commit message, the code comment, the spec,
the CHANGELOG and a test docstring before anyone checked it against a parser.

⚠️ **And twice the test written to guard the rule pinned the wrong answer**, in a docstring citing a
verification that had not happened. That is worse than an unexplained test: it stops the next reader
from checking.

### Findings, by round

| Round | Finding | State |
|---|---|---|
| 6 | **FR-052 stripped Unicode whitespace** → false green on `\xa0` | resolved |
| 6 | The absorb test used substring assertions; **three** corruptions of that splice walked through | resolved — asserts the whole line |
| 6 | A `)` of prose against the deletion boundary had no fixture | resolved |
| 6 | `--write` drops the UTF-8 BOM | **filed #68** |
| 7 | **FR-052 stripped after decoding** → false green *and* false red | resolved |
| 7 | Two tests **pinned** that defect | resolved — both assert the opposite |
| 7 | The six characters of the set were unpinned | resolved |
| 8 | **The delimiter set is asymmetric** → false green on a trailing VT/FF | resolved |
| 8 | The test added in round 7 **pinned** that too | resolved |
| 8 | FR-052 is POSIX-shaped (Windows trims trailing spaces) | **stated in spec + CHANGELOG** |
| 8 | **Balanced parentheses truncated at the first `)`** | **filed #71 — LIVE, 104 occurrences** |

### #71 deserves its own paragraph: it is the first LIVE finding in eight rounds

CommonMark allows balanced parentheses in a destination; `MD_LINK_RE` stops at the first `)`. So
`[…](…/Log Analysis (February - Monthly) - PY02-03.docx)` is cut short and reported dead — while the
file is on disk. **104 in one repository**, all in mirrored attachment filenames.

The report **conceals itself**: its own `(resolves to …)` wrapper supplies the missing parenthesis,
so the truncated path reads as complete and correct. Someone who checks finds the file present and
concludes the *gate* is broken. That repo runs `dangling: warn`, so they print rather than gate —
raising it to `repo`, the direction the rest of the fleet has gone, would close the wall on 104
files that exist.

Measured at 104, not the 5 the review reported: that count was of what the gate *emits* under that
repo's current scope, this one is of the tree.

### CI caught what eight review rounds did not

`b57a08c` turned all four Windows jobs red. Two fixtures were Linux-only: one hard-coded a `/`
separator, the other assumed a file can be named `" "` — Windows trims trailing spaces from a path
component. **Four of these rounds were explicitly hunting platform-independence and neither
surfaced.** The matrix found both in one run. Nothing replaces running it somewhere else.

### Seeding

Rounds 6-8 seeded **20** defects between them. All are caught by the suite today; **five were not**
when the round began — the class-only suffix eater, the `repair` write path, the non-empty-text
suffix eater, re-stripping after decode, and the symmetric edge set. Every one of those five would
have deleted document content or hidden a broken link with the suite fully green.

**274 tests pass** (was 251 on `main`).

Not reproducible in local CI: none. Public repo, Actions run.

txemi:
## Split — superseded by #73, with the rest filed as #74

This PR is being closed **without losing anything**. It carried two changes, and only one of them
was #52:

| | where it went |
|---|---|
| the widening (`[^\]]+` → `[^\]]*`) and its blast-radius guards | **#73**, byte-identical, ready |
| FR-052 — the whitespace *around* a destination | **#74**, with the full nine-round history |

### Why

The first change is two characters and has not moved since round 1. The second was not asked for by
#52: it grew out of a review finding in round 3, and it took **nine rounds**, because *"which
spellings of whitespace delimit a destination"* is a parsing question with at least five dimensions.
Three rounds found a **false green** in the previous round's fix; one traded a false green for five
false reds.

That is not a sign the work was bad — the measurements are sound and every round found something
real. It is a sign the two changes do not belong in one PR: a reviewer of #52 cannot tell which
output change came from which intent, which is the same reason #65, #67, #68 and #71 were filed
rather than folded in.

### What survives here

The nine review ledgers below are the record and they stay put. #74 points at them. Everything they
say about the widening applies unchanged to #73; everything they say about FR-052 is the content of
#74.

The single most transferable finding, stated once more because it cost the most:

> **When a rule is about syntax, adjudicate it against the reference parser — not against the
> requirement sitting next to it.** The rounds that reasoned from the spec were wrong every time.
> The rounds that ran `cmark` were right immediately, and each one overturned a claim the previous
> rounds had already propagated into five documents.

And its corollary, which is why the loop kept failing to notice: **twice, the test written to guard
the rule pinned the wrong answer, in a docstring citing a verification that had not happened.**

PR #70, opened 2026-08-13
gate: the own_web keys, so feature 016 can actually be switched on

txemi:
Feature 016 shipped in **v0.21.0** — in the CLI. **No gate invoked it**, so it protected nothing. This is the wiring.

## What it adds

Both recipes — bash and PowerShell — read three keys and pass them through:

```json
{ "web": true, "own_web": ["your-org"], "own_web_from_origin": true, "own_web_max": 5 }
```

## Three details, each following a rule an existing key already established

- **`own_web_from_origin` is its own key, not a sentinel inside the list.** An owner literally called `origin` has to stay expressible — the same reason the CLI has a flag rather than `--own auto`.
- **A non-numeric budget counts as ABSENT, never as infinite.** Silently *widening* an allowance is the one direction a config typo must not be able to go; `dangling_max` established that.
- **Exit 1 is read as a likely CONFIG error — but only when this run passed an `own_*` flag.** 016 makes exit 1 reachable from configuration (a budget with no owners, an unresolvable origin), and reporting that as a red gate would send someone hunting for broken links that do not exist. It is **not** exclusively a usage error, though: `uvx` exits 1 on its own failures and an uncaught exception exits 1 too. The first draft of this PR swallowed it unconditionally, and that was a regression measured on a repo carrying **no `own_*` keys at all**: `rc=1` on `main`, `rc=0` on the branch, with `fail_closed: true` not rescuing it. So the swallow is guarded by whether a flag was actually passed, and under `fail_closed` it does not become a pass — it becomes **4**, marked final, *not* routed to `bail()` (which exits the script and would skip every axis after this one, the bug this same pass was fixed for in #45).

## The tests assert the invocation, not the verdict

That is not a style choice. Without a token an unreadable destination is `web_unverifiable` and the gate exits 0 **whether or not the flags were passed**, so a verdict-based test cannot tell the difference — and did not: **dropping the `--own` loop entirely left every other recipe test green** until the shim started recording argv. The same held for the web guard itself: *"the keys do nothing when the axis is off"* passed with the guard removed.

## What two adversarial rounds changed

Every item below was found by mutation, not by reading — each is a line whose removal left the suite green:

- The **unconditional exit-1 swallow** above (blocking; measured against `main` across six exit codes × `fail_closed`, now byte-identical in all twelve cells for a repo with no `own_*` keys).
- **`own_web_max: 0` was dropped by the PowerShell recipe**: `CfgOr` tests truthiness and PowerShell reads JSON `0` as `$false`, so the one value whose distinction from *absent* is the entire point of a budget silently became absent. Read by property presence now — and the same fix was needed for `own_web`, where a one-element list unwraps to its element.
- **The `RC_IS_FINAL` immunity had no test.** `web-check`'s codes are all in 0..4 and none of them means *unreachable*, so its `4` — exactly how 016 reports an owned destination with no uuid — must escape the `rc>3` fail-open heuristic. Removing the immunity turned a genuine 4 into **0** under the default, with the whole suite green.
- **Empty owner entries passed silently.** `["" ]` flattens to exactly what an absent key gives, so the list length is now read separately; a partially-empty list is named too (the config listed three owners and enforced one).
- Two warnings existed **only in bash**; both are in the `.ps1` now.
- **Nothing had ever parsed `darnlink-gate.ps1`** — the recipe tests skip on Windows and no job ran `pwsh`. A parse job is added to CI.

**325 tests.** Mutations killed: the three `own_*` flags each arming the swallow on its own, the `web` guard, ignoring `fail_closed`, `4`→`1`, `bail()` instead of a final rc, both empty-entry warnings, the F4 presence read, non-numeric budget acceptance, dropping the explicit `0`, and the `RC_IS_FINAL` immunity.

## Not in this PR

Turning the keys on anywhere. The fleet is at zero owned-without-uuid links today, so the rule would find nothing — its value is keeping that zero, and that is a per-repo decision.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -111,3 +111,22 @@ jobs:
         run: uvx --from . darnlink .
       - name: The exact README one-liner, end-to-end over git (pinned to this commit)
         run: uvx --from git+https://github.com/txemi/darnlink.git@${{ github.sha }} darnlink .
+
+  # The PowerShell recipe is SHIPPED and nothing has ever parsed it: the recipe tests skip on Windows
+  # (`pytestmark`), no job runs `pwsh`, and this PR adds new multi-line constructs to it. Parsing is
+  # not testing — it never runs the gate — but it is the cheapest wall against the one failure mode a
+  # bash-only fleet cannot see: a consumer on Windows getting a script that does not even start.
+  # Its own job on ubuntu (pwsh is preinstalled there), not a step in the 8-cell matrix: the file is
+  # OS-independent, so parsing it once is enough.
+  ps1-syntax:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Parse recipes/darnlink-gate.ps1
+        shell: pwsh
+        run: |
+          $errs = $null
+          [void][System.Management.Automation.Language.Parser]::ParseFile(
+            (Resolve-Path recipes/darnlink-gate.ps1).Path, [ref]$null, [ref]$errs)
+          if ($errs) { $errs | ForEach-Object { Write-Host "::error::$($_.ToString())" }; exit 1 }
+          Write-Host "recipes/darnlink-gate.ps1 parses clean"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,30 @@ All notable changes to darnlink are documented here. The format is based on
 
 ## [Unreleased]
 
+### Added
+- **`darnlink-gate`: the `own_web` keys, so feature 016 can actually be switched on.** The rule
+  shipped in v0.21.0 lives in the CLI; until now no gate invoked it, so it protected nothing. Both
+  recipes — bash and PowerShell — now read `own_web` (a list of owners), `own_web_from_origin` (bool)
+  and `own_web_max` (int) and pass them through.
+
+  Three details keep it honest, and each is the same rule an existing key already follows:
+
+  - **`own_web_from_origin` is its own key, not a sentinel inside the list** — an owner literally
+    called `origin` has to stay expressible, the same reason the CLI has a flag rather than
+    `--own auto`.
+  - **A non-numeric budget counts as ABSENT, never as infinite.** Silently widening an allowance is
+    the one direction a config typo must not be able to go (`dangling_max` established this).
+  - **Exit 1 is a CONFIG error, not a verdict about the repository.** Feature 016 makes it reachable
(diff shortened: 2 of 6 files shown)


txemi:
## Verdict — two adversarial rounds, closed by the third criterion

Copilot's quota is exhausted account-wide, so both rounds were run by a review subsession under the
same contract: measure, don't read; try to refute; report what you looked at and found clean.

| Ronda | Bloqueantes | Mayores | Menores | Cierre |
|---|---|---|---|---|
| 1 | **1** — the exit-1 swallow was unconditional | 3 | 4 | applied, all mutation-verified |
| 2 | 0 | 2 | 4 | applied; the regression sweep **refuted** |

### The one that mattered

Round 1's blocker was a **regression for consumers who never adopt this PR**: a repo carrying no
`own_*` keys went `rc=1` on `main` and `rc=0` on the branch, and `fail_closed: true` did not rescue
it. Round 2 repeated the sweep as its main attack — repo with no keys, `web-check` forced to
0/1/2/3/4/127, with and without `fail_closed` — and got **12 of 12 cells byte-identical to `main`**
in rc, stdout and stderr. With `own_web` set, only the two intended cells differ.

### What round 2 found that reading had not

Every item was a line whose removal left the entire suite green:

- **The exit-4 immunity had no test.** Remove it and a genuine 4 — how 016 reports an owned
  destination with no uuid — becomes **0** under the default, which is what pre-commit and pre-push
  run. Two tests now: the 4 survives, and it is *carried* rather than `bail()`'d, so the later axes
  still speak. Asserting the code alone cannot tell those apart — both give 4.
- **Only `--own-max` was pinned to arming the exit-1 reading.** `--own` and `--own-from-origin` each
  survived a mutation.
- **`[""]` was a silent no-op**, indistinguishable from an absent key once flattened. And in
  PowerShell it was worse: a one-element array unwraps, so `CfgOr` read it as absent.
- **Nothing had ever parsed `darnlink-gate.ps1`** — recipe tests skip on Windows, no job ran `pwsh`.

### Why this closes rather than loops again

Criterion 1 of the three (**dry**): round 2's remaining items were coverage gaps and documentation,
not behaviour — and the behavioural attack it led with was refuted by measurement rather than by
argument. The residual window it named (an exit 1 from `uvx` itself in a repo that *did* adopt 016
still goes green under fail-open, where filtering on the `error: --own` stderr line would close it)
is recorded as a follow-up, not applied: it changes how the run's streams are handled, which is a
larger change than the one it protects.

**325 tests. Eight mutations this round, all killed.** CI green, `mergeStateStatus: CLEAN`.

PR #83, opened 2026-08-16
Absolute local paths are now a named finding, not an unchecked blind spot

txemi:
## Summary

`[x](/home/user/notes.md)` currently receives no check of any kind. `is_local_relative` excludes
anything starting with `/`, so such a link is invisible to both `dangling` (shares the exclusion)
and `out_of_scope` (that axis names a real location outside `--root`; an absolute path names none).
Unlike a relative link — which resolves to "this repo" on any clone — an absolute path can only
ever resolve on the one machine that wrote it, and until now, silently.

- New `Kind.ABSOLUTE_LOCAL_PATH`, detected in `robustify.py` alongside the existing dangling check
  (same `is_local_relative` gap, same FR-047 encoded-href handling).
- Surfaced in `darnlink check` as its own axis (`absolute_local_path` in `--json`, an
  `[absolute-local-path]` line in text output) — same treatment as `dangling` (FR-049): report-only,
  **deliberately absent from the exit code**, so no consumer's gate goes red the day it upgrades.
- 12 new tests in `tests/test_absolute_local_links.py`, mirroring `test_dangling_links.py`'s style,
  including the case this is easiest to confuse with an existing axis
  (`test_out_of_scope_relative_link_is_not_absolute_local_path`).

## Why

Grew out of auditing whether a gate hardening (fail on any local link that escapes the repo root)
would be safe to turn on across the repos that run `darnlink-gate.json`. Absolute local paths turned
out to be a separate, pre-existing blind spot with near-zero adoption cost — verified by running
`darnlink check --json` against the 9 real repos currently gated by darnlink: every one reports
`exit_code: 0` unchanged, with `absolute_local_path.count` ranging from 0 to 84 (mostly
`/wiki/...`/`/media/...`-style paths baked into Confluence/SharePoint mirror exports — legitimate
things to *see*, not things that should ever fail a build).

## Test plan
- [x] `uv run pytest` — 397 passed (full suite, including the 12 new tests)
- [x] `bash tools/check.sh` — darnlang, tests, repair/robustify/dangling gate all green
- [x] `darnlink check .` on darnlink's own tree — clean
- [x] `darnlink check --json` against 9 external repos that run `darnlink-gate.json` — exit code
      unchanged at 0 in all of them

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,16 @@ All notable changes to darnlink are documented here. The format is based on
   so the milliseconds belong to the machine and are not quoted as a property of the change.
 
 ### Added
+- **A plain link written as an absolute filesystem path (`/home/user/x.md`) is now a named finding
+  (`absolute_local_path`), instead of receiving no check at all.** `is_local_relative` excludes
+  anything starting with `/`, so such a link was invisible to every existing axis: not `dangling`
+  (that one shares the exclusion), not `out_of_scope` (that one names a real location outside
+  `--root`; an absolute path names none). Unlike a relative link — which resolves to "this repo" on
+  any clone — an absolute path can only ever resolve on the one machine that wrote it, and silently,
+  since it was never reported either. Same treatment as `dangling` (FR-049): report-only, its own
+  axis in `check --json`, and deliberately absent from the exit code, so no consumer's gate goes red
+  on the day it upgrades.
+
 - **`tests/test_recipe_examples.py` — the examples are code, and nothing had ever run them.** The
   release that added a `ps1-syntax` job on exactly that argument then put non-trivial shell into two
   example files with no gate at all. The tests **extract** the commands from the example files and
--- a/src/darnlink/cli.py
+++ b/src/darnlink/cli.py
@@ -185,6 +185,12 @@ def _run_check(root: Path, excludes: set, as_json: bool, block_markers: tuple,
     # gate is the caller's policy; the core just names what it sees.
     dangling = [f for f in rob.findings if f.kind is Kind.DANGLING]
 
+    # Absolute-local-path axis (017): plain links written as an absolute filesystem path. Same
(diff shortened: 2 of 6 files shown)


txemi:
## Substitute review (Copilot quota exhausted)

Acting as reviewer-in-loop substitute per the PR workflow. Verified against worst-case failure modes (silent gate red, exit-code change, behavioral regression).

### ✅ Verified safe — no blocking issues
- **Exit code never changes.** Code read + live test: `darnlink check` with and without absolute-path findings returns 0 in both text and JSON modes.
- **Phase B loop unaffected** — DANGLING and self-link branches untouched, anchorable links unaffected.
- **No OUT_OF_SCOPE collision** — absolute paths (start with `/`) can't enter the relative-path codepath by construction.
- **Kind wiring consistent** across codebase.
- **Full test suite:** 397 passed, no regressions.
- **Consumer script** (`recipes/darnlink-gate`) literally cannot see the new axis yet (parses only `dangling` key).

### ℹ️ Low-severity findings (no action needed; documented below)

1. **Coverage gap (inherited, not regressed):** Windows drive-letter (`C:\...`) and tilde-relative (`~/...`) paths fall through to dangling-check instead, because `is_absolute_local_path` recognizes only POSIX leading `/`. Identical behavior pre-exists for `is_local_relative`. Suggest one-line docstring caveat.

2. **Angle-bracket destinations (inherited):** `[a](<...>)` destinations aren't stripped, so they miss both this and dangling checks. Pre-existing; this PR is consistent.

3. **Robustify text output omits finding** — only `check` mentions it. Consistent with `Kind.DANGLING` (informational axes are check-only).

4. **Missing spec document** — features 001–016 have `specs/0NN-.../spec.md`. This one doesn't. No functional impact, but breaks convention.

### Conclusion
No findings rise to "fix before merge." The core safety guarantee (no consumer goes red) is verified by reading + running. Ready for merge once CI stays green.

PR #84, opened 2026-08-16
docs: share one instruction file across agents via symlinks

txemi:
AGENTS.md and .github/copilot-instructions.md become symlinks to CLAUDE.md, so Copilot — and any agent following the AGENTS.md convention — reads the same conventions as Claude, with no second copy that can drift.

Git stores symlinks natively (mode 120000), so they travel on clone; no build step and no generator involved.

Verified: both resolve to CLAUDE.md, and git records them as 120000.

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -1,0 +1,1 @@
+../CLAUDE.md
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,1 @@
+CLAUDE.md


txemi:
## Review verdict — round 1

Copilot could not review this PR (account quota exhausted), so an adversarial Claude sub-session reviewed it instead, per the project's substitute-mechanism rule.

| Check | Result |
|---|---|
| Symlinks stored natively | `git ls-files -s` → mode `120000` for both; `realpath` resolves to `CLAUDE.md` |
| Link gate (dogfood) | `darnlink .` → 4 uuids, **0 duplicates**. `CLAUDE.md` has no frontmatter, so all three paths are ignored for the index — no double counting |
| Link gate (MAX profile) | 0 changes, exit 0 |
| Tests | 397 passed |
| Language gate | 0 legacy lines |
| Wheel packaging | `packages = ["src/darnlink"]` → neither file ships in the wheel |
| Sdist packaging | Both included, serialized as tar symlink entries (`type='2'`); `twine check` PASSED |
| Windows | CI `test`+`smoke` green on `windows-latest` for this branch |
| CI overall | 25/25 checks green |

**Not verified:** whether Copilot follows the symlink when reading `.github/copilot-instructions.md`, or reads the literal link target text. Cannot be determined from a local environment — flagged as unverified rather than assumed either way.

**Latent note, not blocking:** darnlink's write operators (`frontmatter_edit.write_text_keep_newlines`, used by `repair.py`/`robustify.py`) do not check `path.is_symlink()` before opening for write. Not triggered today (measured: 0 changes), but if something ever links to `AGENTS.md` as a target and the gate decides to create frontmatter for it, the write would go **through the symlink into `CLAUDE.md`**.

**Verdict: mergeable as-is.**

PR #85, opened 2026-08-16
fix: a symlink is another name for a file, not another file

txemi:
`iter_markdown_files` walked every path and yielded it as-is, so a symlink to a `.md` **inside the scanned root** was read as a separate document.

Sharing one instruction file across agents — `AGENTS.md`, `.github/copilot-instructions.md` and `CLAUDE.md` pointing at one source, which is standard practice — therefore made the same `uuid` appear at three paths, and the integrity check failed with *"uuid in multiple files"*. That is a **false positive**: a symlink creates no new entity, only a new name.

**Measured on a real repo (2026-08-16):** adopting that layout turned the gate red and blocked every push across the fleet until the links were reverted.

```
[integrity] repair: 1 | unresolved: 6 -> FAIL
uuid e1080ad2-... in multiple files; left untouched   (x6)
-> exit 2 (integrity failure)
```

Same tree, after this change: `indexed uuids: 2 | duplicate uuids: 0`.

## What changed

Files are deduplicated by resolved path, and the **canonical** path is the one yielded.

That second half is not cosmetic. Relative links in the body resolve against the directory of the file they were read from, so reporting `.github/copilot-instructions.md` made a body link `inventory/notes/` look broken and repair wanted to rewrite it to `../inventory/notes/`. Walk order would otherwise decide which name wins — `AGENTS.md` sorts before `CLAUDE.md`.

## Decisions worth reviewing

- **Symlink out of the root** → skipped, deliberately. The scan is defined by the root it was given; following a link out of it would index files the caller never asked for. Mirrors how out-of-root link *targets* are reported rather than chased.
- **Broken symlink** → skipped quietly. It is still a valid link target, and reporting a dangling target belongs to the dangling gate, not the indexer.
- **Link as a target** (`](AGENTS.md)`) still resolves; only the index is deduplicated.

## Tests

6 new in `tests/test_symlink_dedup.py`: dedup, the three-name regression, out-of-root, broken link, walk order, and the subdirectory case that shifted link resolution. Full suite: **403 passed**.

## Not addressed here

Write operators (`repair.py`, `robustify.py`) still do not check `is_symlink()` before opening for write — if the gate ever created frontmatter on a link, it would write **through** it into the shared source. Same family, separate change.

### DIFF
--- a/src/darnlink/cli.py
+++ b/src/darnlink/cli.py
@@ -38,6 +38,7 @@ def _findings_json(
     link_ignored: Optional[List[Path]] = None,
     suppressed: int = 0,
     only: Optional[set] = None,
+    out_of_root: Optional[List[Path]] = None,
 ) -> str:
     return json.dumps(
         {
@@ -49,6 +50,9 @@ def _findings_json(
             # feature 006: opted out as a SOURCE only — still indexed as a target
             "link_ignored_files": [str(p) for p in (link_ignored or [])],
             "invalid_frontmatter_files": [str(p) for p in (invalid or [])],
+            # Not decoration: a gate consumes the JSON, not the human text. Omitting it here
+            # would fix the report only where nobody automated is looking.
+            "out_of_root_links": [str(p) for p in (out_of_root or [])],
             "findings": [{"kind": f.kind.value, "file": str(f.file), "detail": f.detail} for f in findings],
         },
         indent=2,
@@ -66,7 +70,7 @@ def _run_repair(root: Path, write: bool, excludes: set, as_json: bool, block_mar
 
     if as_json:
         print(_findings_json(result.findings, wrote, write, result.ignored, index.invalid,
-                             result.link_ignored, result.suppressed, only))
+                             result.link_ignored, result.suppressed, only, index.out_of_root))
     else:
         print(f"darnlink repair — root: {root}")
         if only is not None:
@@ -118,7 +122,7 @@ def _run_robustify(root: Path, write: bool, create_frontmatter: bool, excludes: 
 
     if as_json:
         print(_findings_json(result.findings, wrote, write, result.ignored, result.invalid,
-                             result.link_ignored, result.suppressed, only))
+                             result.link_ignored, result.suppressed, only, result.out_of_root))
     else:
         print(f"darnlink robustify — root: {root}")
         if only is not None:
@@ -130,6 +134,9 @@ def _run_robustify(root: Path, write: bool, create_frontmatter: bool, excludes: 
             print(f"  [create-readme] {f.file}: {f.detail}")
         for f in skipped:
             print(f"  [no-frontmatter] {f.file}: {f.detail} (use --create-frontmatter to allow)")
+        for p_ in result.out_of_root:
+            print(f"  [out-of-root-link] {p_}: symlink whose target lives outside the scanned root; "
+                  f"not indexed — its uuid will NOT resolve (widen the root, or replace the link with a copy)")
         for f in out_of_scope:
             print(f"  [out-of-scope] {f.file}: {f.detail}")
         for f in target_writes:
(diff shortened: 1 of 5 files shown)


txemi:
## Review verdict — round 1: *mergeable with changes* → changes applied

Copilot could not review (account quota exhausted), so an adversarial Claude sub-session reviewed it. It measured everything against real trees rather than reasoning about it, and it found a real hole **in this PR's own claim**.

| # | Finding | Action |
|---|---|---|
| 1 | **Out-of-root symlinks were silenced.** The PR said it "mirrors how out-of-root link targets are reported" — it reported nothing. Such a file *used* to be indexed (reading a symlink follows it), so its uuid resolved; skipping it silently turns a working robust link into `unresolvable` with nothing naming the cause | **Fixed** in `b2039f7`: collected in `FrontmatterIndex.out_of_root` and printed by the CLI with the cause and both ways out. New test asserts the report, not just the skip |
| 2 | **No consumer breaks.** All 4 callers already `.resolve()` before comparing; `resolve()` is idempotent | Confirmed, no change needed |
| 3 | **`resolve(strict=True)` hides nothing.** A `chmod 000` file still resolves (stat, not read); circular symlinks are caught by `OSError` (ELOOP); broken links behaved the same before | Confirmed, no change needed |
| 4 | **Cost: ~4×** — 3533 files went 0.18s → 0.7s | Accepted. Sub-second at this scale, ~3-4s at 18k files, once per run. Worth it to stop a false `exit 2` |
| 5 | **Hardlinks still duplicate.** `resolve()` collapses symlinks, not inodes: two hardlinks to one file still report "uuid in multiple files" | Out of scope here (see below), but it is the same false positive by another route |

Organic validation worth noting: run against a real 3533-file tree, the new code found **3533 vs 3534** — a genuine pre-existing symlink that was being indexed twice and is now correctly collapsed to its canonical path.

Full suite after the fix: **404 passed**.

## Still not addressed here

- **Hardlinks** (finding 5). Same class of false positive, different mechanism — `os.stat().st_ino` rather than `resolve()`. Deliberately separate: it changes the dedup key for every file, not just links.
- **Write operators** (`repair.py`, `robustify.py`) still do not check `is_symlink()` before opening for write.

txemi:
## Review verdict — rounds 2 and 3: *do not merge* → both blockers fixed

| Round | Blocker | Fix |
|---|---|---|
| 2 | The out-of-root warning reached **1 of 6** output paths. `--json`, `--robustify` (both modes) and `check` — the subcommand documented for CI, and the one a gate actually consumes — were silent | `a1985b6`: wired through `_findings_json`, `_run_robustify` (own field on `RobustifyResult`, since `plan_robustify` never builds an index) and `_run_check` |
| 3 | **`web-check` was still mute** in both branches (`cli.py` offline listing, `weblinks.py:check_web_links_online`). The commit had claimed "ALL output paths" | `1651071`: both branches wired, text and JSON |

**The pattern is the finding.** Three times in this PR completeness was *declared* instead of *enumerated* — the same habit the change exists to break. So the test no longer asserts a total: it drives **eight explicitly listed argv sets** through `main()`, and a surface added later fails loudly instead of quietly falling outside a claim.

Verified by running each command against a real fixture, not by trusting the tests:

```
✅ darnlink ROOT              ✅ darnlink ROOT --json
✅ darnlink ROOT --robustify  ✅ darnlink ROOT --robustify --json
✅ darnlink check ROOT        ✅ darnlink check ROOT --json
✅ darnlink web-check ROOT    ✅ darnlink web-check ROOT --json
```

All three JSON payloads parse with `json.tool` and carry `out_of_root_links`.

### Confirmed by round 3, no change needed

- **Exit code stays untouched, and the premise holds**: a robust link to the lost uuid already fails as `unresolvable` with `exit 2` (measured). An out-of-root symlink with no inbound links exits 0 — harmless, so failing on it would break trees that are fine.
- **No double reporting**: `_run_check` builds one index and `plan_robustify`'s own `out_of_root` is never printed.

### Also fixed

Two English comments had Spanish clauses spliced in. The language gate caught the second one — working as intended. The new `web-check` lines carry the same `-> informational (does not affect the exit code)` suffix as their siblings.

### Still out of scope, stated rather than implied

- **Hardlinks**: `resolve()` collapses symlinks, not inodes, so two hardlinks to one file still report "uuid in multiple files". Same false positive, different mechanism; it would change the dedup key for every file.
- **Write operators** still do not check `is_symlink()` before opening for write.

Suite: **405 passed**.

PR #88, opened 2026-08-16
D1a: an anchor comment stops separating a link from its pandoc attrs (#65)

### LINKED ISSUE #65: --write detaches a pandoc attribute block: the anchor lands between the link and its {…}
## What happens

When a link carries a pandoc attribute block, `--write` inserts the uuid anchor **between the link
and its attributes**, detaching them:

```
before:  [](B.md){.cls}
after:   [](B.md) <!-- uuid: bbbb… -->{.cls}
```

Pandoc requires the attribute block to follow the closing `)` **immediately**. With the comment in
between it no longer applies, so the rendered output silently loses the class, the width, the height
— whatever the block carried. The document still renders; it just renders wrong.

## Reproduction

```python
# A.md:  ![](B.md){width="1.1in" height="2in"}       B.md has a uuid
apply_robustify(plan_robustify(root))
# A.md becomes:
#   ![](B.md) <!-- uuid: … -->{width="1.1in" height="2in"}
```

## This is pre-existing, not new

Measured on both shapes — it behaves identically with a non-empty text (`[x](B.md){.cls}`), so it
predates the empty-text widening and has presumably always been there.

**What is new is the exposure.** That widening routes empty-alt embeds into the write path for the
first time, and the corpus that motivated it is converted `.docx`/`.odt` documents, where the
attribute suffix is on essentially every image. So the population of links that can hit this grows
substantially even though the defect itself is unchanged.

Frequency today: **0** across thirteen local repositories — every empty-text link with a live target
points at a `.jpg`/`.jpeg`, which is unanchorable, so nothing reaches the write path yet. Latent,
not live. That will stop being true the first time one of those documents links to a `.md`.

## Why it was not fixed in the PR that found it

It was found while reviewing a regression fix. Folding a behaviour change into that fix would have
made the fix unreviewable — the reviewer could no longer tell which output change came from which
intent. Declared in the spec's blast-radius section and filed here instead.

## Suggested direction

The anchor's insertion point is the end of the link's match span. The span deliberately stops at the
`)` — that is what keeps the suffix out of the *match*, which is correct and must stay. What needs
to change is where the **edit** lands: after any attribute block that immediately follows, not at the
end of the span.

Watch out for the interaction with `_TRAILING_UUID_RE`, which decides whether a link is already
robust by matching from the end of the span. If the anchor moves past a `{…}`, that detection has to
move with it, or a link anchored once will be seen as plain and anchored again.

Whatever the fix, it needs a **write** test: detection reports an identical finding either way, so
this class of defect is invisible to every test that only inspects findings. That is exactly how it
survived to now.

txemi:
Fixes #65. Part of the darnlink v0.24.0 block (D1-D7).

`{width="1.1in"}` anchored by `--write` used to become
` <!-- uuid: ... -->{width="1.1in"}` — the comment landing between the link and the
attribute block pandoc requires immediately after `)`. The block still renders, silently ignored.

Reproduced first (exactly the issue's example), then fixed at 5 sites — every consumer of the link
grammar, enumerated before touching any of them:

- `ROBUST_LINK_RE` — optional `attrs` group
- `RobustLink` — carries `attrs`
- `find_plain_links` / `find_detached_anchors` — skip past an attrs block before checking for a
  trailing anchor (otherwise an anchored link reads as plain again on the NEXT run and gets a
  second anchor)
- `emit_robust_link` — attrs go right after `)`, no space
- `robustify.py` write site + `repair.py` — both write paths now round-trip attrs

Two existing tests asserted the old (buggy) shape as correct; rewritten to assert the fix, keeping
the historical reasoning for why the old restriction existed.

**407 tests pass** (were 405; +2 net, 2 rewritten). darnlang clean, repair/robustify/dangling gates
clean on this tree.

### DIFF
--- a/src/darnlink/links.py
+++ b/src/darnlink/links.py
@@ -10,6 +10,20 @@ import re
 from dataclasses import dataclass
 from typing import List, Sequence, Tuple
 
+# The BODY of a pandoc attribute block, between its own `{` and `}` — one definition, shared by
+# ROBUST_LINK_RE's `attrs` group below and by `_PANDOC_ATTR_RE` further down. `[^{}\n]*` alone (the
+# first version of this pattern) treats a `}` inside a quoted attribute value as the block's own
+# closing brace and cuts the match short there — not a miss, a WRONG match, and a corrupting one
+# once something is spliced at the point it stopped. `"[^"\n]*"` makes a quoted run atomic, so `{`
+# and `}` inside a quote no longer terminate the block early; an unterminated quote still fails the
+# whole match, same as an unbalanced `{` always did. Kept as ONE definition on purpose — two
+# independent copies of "what an attrs block looks like" is exactly the shape that drifted before
+# (`is_local_relative`'s four call sites, FR-052), and it already drifted once here: an earlier
+# revision of this fix updated only `_PANDOC_ATTR_RE` and left this literal on the old, corrupting
+# pattern, so `repair` silently stopped recognising (and therefore stopped fixing) an already-correct
+# robust link whose attrs held a quoted `}` — no finding, no write, the stale path just stayed stale.
+_PANDOC_ATTR_BODY = r'(?:[^{}"\n]|"[^"\n]*")*'
+
 # A robust link: a Markdown link immediately followed (any whitespace) by a uuid HTML comment.
 # The optional `attrs` group is a pandoc attribute block (`{.cls}`, `{width="1in"}`, …): pandoc
 # requires it immediately after the link's `)`, with no whitespace, so it must be matched BEFORE
@@ -17,7 +31,7 @@ from typing import List, Sequence, Tuple
 # right up to `{`, misparsing `[x](y){.cls}` as `[x](y)` followed by unrelated `{.cls}` text.
 ROBUST_LINK_RE = re.compile(
     r"\[(?P<text>[^\]]*)\]\((?P<href>(?:[^()\s]|\((?:[^()\s]|\([^()\s]*\))*\))+|[^)]+)\)"
-    r"(?P<attrs>\{[^{}\n]*\})?\s*<!--\s*uuid:\s*(?P<uuid>[0-9a-fA-F-]{36})\s*-->"
+    r"(?P<attrs>\{" + _PANDOC_ATTR_BODY + r"\})?\s*<!--\s*uuid:\s*(?P<uuid>[0-9a-fA-F-]{36})\s*-->"
 )
 # Any inline Markdown link. `text` is `*`, not `+`: `[](dest)` is a link with empty text, and the
 # empty alt of `![](dest)` is what pandoc emits for every image in a converted .docx/.odt. Requiring
@@ -81,7 +95,9 @@ _TRAILING_UUID_RE = re.compile(r"\s*<!--\s*uuid:\s*[0-9a-fA-F-]{36}\s*-->")
 # rule -- a block after a space is not attached to anything and pandoc ignores it). Matched with
 # .match(content, pos) the same way as _TRAILING_UUID_RE, so callers can skip past it before
 # checking for the anchor comment -- see `_skip_attrs` below for why that skip has to exist at all.
-_PANDOC_ATTR_RE = re.compile(r"\{[^{}\n]*\}")
+# Shares `_PANDOC_ATTR_BODY` with `ROBUST_LINK_RE`'s own `attrs` group (see that constant's comment
+# for why one definition, not two, matters here specifically).
+_PANDOC_ATTR_RE = re.compile(r"\{" + _PANDOC_ATTR_BODY + r"\}")
 
 
 def pandoc_attrs_at(content: str, pos: int) -> str:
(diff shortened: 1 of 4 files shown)


txemi:
## Adversarial review (Copilot substitute — quota exhausted: "Copilot was unable to review this pull request because the user who requested the review has reached their quota limit.")

**Round 1** — full-file read of `links.py`/`robustify.py`/`repair.py` (not just the diff), scripted reproduction of edge cases, seeded-defect check.

Areas examined and found **clean**:
- `ROBUST_LINK_RE`'s ordering (`attrs` group matched before the anchor's `\s*`) — correct, verified `[x](y){.cls}` is not misparsed.
- `find_plain_links` / `find_detached_anchors` skipping past `attrs` via `_skip_attrs` before checking for a trailing anchor — correct; a link this tool just anchored does not get re-detected as plain on a second pass (existing test `test_65_a_second_pass_over_an_attrs_anchored_link_is_a_no_op`, reconfirmed).
- Duplication risk (same uuid anchored twice) — ran two full passes over several fixtures; no duplication in any case tried.
- `repair.py` threading `link.attrs` through `emit_robust_link` — correct for the common shapes (existing tests `test_an_attribute_block_before_the_anchor_IS_a_robust_link`, `test_repair_write_leaves_a_pandoc_attribute_suffix_untouched`).
- Two links on the same line, one with `attrs` one without, same target uuid — both anchored independently, correct output.
- `{data-x="a{b}"}` (unquoted nested `{`) — regex fails to match at all (by design, undocumented depth bound), falls back to the *old* pre-#65 placement (misplaces but does not corrupt) — no worse than `main`.
- The absorb/`_hspace_start` boundary next to an attrs block and next to prose parens — existing tests `test_absorbing_a_stray_anchor_does_not_eat_an_attribute_block` / `test_absorbing_a_stray_does_not_eat_a_closing_paren_of_prose`, reconfirmed.
- Performance: 3000-line adversarial fixture (many quotes/braces near links) — 0.62s, no pathological backtracking.

**Two real findings, both applied and pushed** (commit `1d80c06`):

1. **Write-side corruption, worse than #65 itself.** `_PANDOC_ATTR_RE`'s body `[^{}\n]*` treats a `}` *inside a quoted attribute value* as the block's own closing brace and cuts the match short there — not a miss, a **wrong** match. On `--write` this splices the anchor comment into the middle of the attribute text: `{data-x="a}b"}` became `{data-x="a} <!-- uuid: … -->b"}`. Reproduced directly against `apply_robustify`; confirmed **absent on `main`** (main only misplaces the intact block, never splits it — this PR's fix made this one case strictly worse than the bug it was fixing). Fixed: attrs body is now `(?:[^{}"\n]|"[^"\n]*")*`, treating a quoted run as atomic. New test: `test_65_a_quoted_attr_value_containing_a_brace_is_not_split`.
2. **Read-side drift, found while fixing #1.** `ROBUST_LINK_RE`'s own `attrs` group was a *second, independent copy* of the same pattern text, not sharing a definition with `_PANDOC_ATTR_RE` — exactly the shape the code's own `pandoc_attrs_at` docstring warns about (`is_local_relative`'s four call sites, FR-052). Hardening only #1 left this one on the old pattern: `find_robust_links` (which `repair` is built on) then failed to match an already-correctly-anchored link whose attrs held a quoted `}` **at all**, so `plan_repairs` reported **zero findings** for a link whose target had genuinely moved — the stale path would have survived silently. Fixed by extracting one shared `_PANDOC_ATTR_BODY` constant used by both regexes. New test: `test_65_repair_finds_a_stale_link_whose_attrs_hold_a_quoted_brace`.

**Seeded-defect check** (both reverted after confirming):
- Commented out the span-widening `link.end + len(attrs)` in `robustify.py` → caught by 3 existing tests (`test_absorbing_a_stray_anchor_does_not_eat_an_attribute_block`, `test_65_the_reported_example_anchors_with_attrs_in_place`, and the new `test_65_a_quoted_attr_value_containing_a_brace_is_not_split`).
- Reverted the `ROBUST_LINK_RE`/`_PANDOC_ATTR_RE` unification (finding #2 above) → caught by the new `test_65_repair_finds_a_stale_link_whose_attrs_hold_a_quoted_brace`.

Both seeded defects were caught without needing further new tests beyond the two added for the real findings — coverage gap closed as part of applying the fix, not left open.

**Gates run**: full suite `uv run pytest -q` — 409 passed (was 407 before this round's 2 new tests). `darnlink .` (repair self-check) and `darnlink . --robustify --create-frontmatter --no-create-frontmatter-for README.md` (MAX self-check) — both clean. `darnlang check --ext all` — clean. `python3 tools/dangling_gate.py` — clean. All via the repo's own `tools/check.sh`, run locally through the pre-commit/pre-push hooks (not bypassed).

**Not reproduced locally**: the Windows leg of the CI matrix and `smoke`/`ps1-syntax` jobs — but CI has since run on the pushed commit and all of those are green (`test (windows-latest, 3.10..3.13)`, `smoke (ubuntu-latest)`, `smoke (windows-latest)`, `ps1-syntax`, `lang`, `title-and-body` all COMPLETED SUCCESS at the time of this comment; the ubuntu `test` matrix was still finishing).

**Verdict**: `listo-salvo-merge`. One round found two real, fixable issues; both applied, tested (including via deliberate seeded defects), and pushed. A second, shorter pass focused specifically on the two fixes just applied (perf stress test, re-grep for any other copy of the old pattern) found nothing further. Closing on gate coverage: the test suite together with the repo's own dogfooding gates (`darnlink`, `darnlang`, dangling) now covers this change end to end, and I verified that coverage directly rather than assuming it.

Issue #74, opened 2026-08-13
FR-052: the whitespace delimiting a destination — a scanner problem, not a strip (split out of #62)


txemi:
Split out of #62 / PR #73, with the full history, because it turned a one-character regression fix
into nine review rounds and it is not the same problem.

## The rule

CommonMark does not count the whitespace *delimiting* a link destination as part of it:
`[x]( B.md )` denotes `B.md`. darnlink's `dangling` axis judged the raw spelling, so such a link
resolved to `dir/ B.md ` and was **reported dead while the file sat right there** — a false red on a
live file, which is how a gate gets switched off rather than fixed.

That much is simple. What follows is why it is not.

## Five dimensions, each of which looked closed until the next round measured it

| # | The question | What was got wrong |
|---|---|---|
| 1 | literal whitespace | the strip was placed *after* percent-decoding |
| 2 | where the rule is applied | computed for the emptiness test, then a **different string** resolved |
| 3 | ASCII vs Unicode | `str.strip()` also eats NBSP and the ideographic space, which CommonMark keeps |
| 4 | encoded vs literal | `%20a.md` denotes `" a.md"`; stripping it made a link to a **missing** file resolve to an existing neighbour |
| 5 | the two edges | not merely different *sets* — the trailing edge is a **scan terminator**, not a strip |

**Three of those were false greens.** One (#5, as first fixed) traded a false green for **five false
reds**. Every one was found by the round *after* the one that "closed" the dimension before it.

### The thing worth keeping from all of it

The rounds that reasoned from the spec — or worse, from the requirement sitting next to this one —
got it wrong every time. The rounds that ran `cmark` got it right immediately, and each overturned a
claim the previous rounds had propagated into five documents. The clause that licensed the first
error was borrowed from FR-047, where it is true:

> *"those spellings denote the same destination"*

True of a revealed scheme or absolute path. **False of whitespace.**

⚠️ And twice, the test written to guard the rule **pinned the wrong answer**, in a docstring citing a
verification that had not happened. That is worse than an unexplained test: it stops the next reader
from checking.

## The formulation that actually matches the parser

Measured over every ordered pair of whitespace characters at each edge — 121 inputs `cmark` parses
as links:

| implementation | agrees |
|---|---|
| symmetric strip of six characters | 93 / 121 |
| `lstrip(6)` + `rstrip(4)` | 109 / 121 |
[...]

PR #89, opened 2026-08-16
D1b: a trailing space no longer turns a file link into an unhealable CONFLICT (#67)

### LINKED ISSUE #67: A trailing space in a destination makes repair emit a CONFLICT it cannot heal
FR-052 (shipped in the `dangling` axis) strips the whitespace around a link destination before
resolving it, because CommonMark does not count that whitespace as part of the destination. Three
other sites resolve the same href **without** stripping, so one link gets two different verdicts
depending on which axis looks at it.

## The three sites

| Site | Consequence of not stripping |
|---|---|
| `names_md` (via `repair`) | **A wrong, unhealable verdict.** See below — this is the one that matters. |
| `_anchor_target` | `check` reports "nothing to anchor" over a link it could anchor |
| `_dir_link_missing_readme` | `--create-readme` does not consider the directory |

Also worth noting for consistency: `is_local_relative` is called on the raw `href` at the top of
`_dangling_target` but on the stripped path in the encoded branch, so ` /etc/nope.md` (absolute, with
a leading space) is reported when FR-046 says an absolute path never is. And `is_web_href` in
`paths.py` **does** strip — the module is internally inconsistent about this already.

## The `repair` case, which is not cosmetic

```
href='old/B.md'    -> repair:  old/B.md -> new/B.md            rewritten
href='old/B.md '   -> conflict: "…is a directory link, but uuid … lives in
                                 …/new/B.md (not a README.md) — path and uuid
                                 disagree; left untouched"      NOT rewritten
```

`names_md("old/B.md ")` is false — the string ends in a space — so the link is classified as a
**directory** link, its uuid is then found in a non-README, and the result is a `CONFLICT`.
`conflict` feeds `integrity_fail`, so the gate **exits 2**. Unlike a `repair`, `--write` never heals
it: the build stays red permanently, and the diagnosis printed is *false* — the path and the uuid
agree perfectly; only the trailing space confused the classifier.

That is the worst shape a gate finding can take: red, unfixable by the tool that reported it, and
explained wrongly.

## Frequency

**0 occurrences across thirteen local repositories** — 0 of 10.116 robust links, 0 of 22.149 plain
links. Latent, not live. It stays latent only as long as nobody hand-writes a link with a stray
space, or pastes one out of a document converter.

## Why it was not fixed where it was found

It surfaced during review of a regression fix (#52 / PR #62). Fixing it touches the `repair`,
`robustify` and `create-readme` axes, none of which that PR owns, and a fix that changes behaviour on
three axes cannot be reviewed as part of a fix for a fourth. FR-052's normative sentence was narrowed
to say "within the `dangling` axis" so the gap is a recorded decision rather than a silence.

## Suggested direction

One shared helper — "the destination this href denotes", doing `split_fragment` → `unquote` →
`strip` — used by every site that resolves an href, rather than four independent spellings of the
same intent. That is the actual defect: the rule is right, and it lives in one function instead of in
the notion of "destination".

Each site needs its own test. Seeded checks confirm that **none** of the current tests pin the
unstripped behaviour, so all three can be changed today without a single test noticing — which is
also why this is worth doing deliberately rather than opportunistically.

---

## Re-scoped (2026-08-13)

The title and the opening of this issue describe a **divergence** between axes — `dangling` strips,
`repair`/`robustify`/`--create-readme` do not. That was true only on the branch of the closed PR #62.
That branch is gone: the whitespace rule is deferred to #74, and on `main` **no axis strips**, so
there is no divergence to report.

**What is still true, and is the part worth keeping**, is the second half: a trailing space makes
`names_md` false, the link is classified as a *directory* link, and `repair` emits a `CONFLICT`
diagnosed as *"path and uuid disagree"* — which is untrue, and `--write` never heals it, so the gate
exits 2 permanently. That reproduces on `main` and is unaffected by any of this.

Read this issue as being about **that**, and read #74 for the stripping rule itself.

txemi:
Fixes #67. Part of the darnlink v0.24.0 block (D1-D7).

`names_md("old/B.md ")` returned `False` — a trailing space defeated `.endswith(".md")` — so
`repair` misclassified an ordinary file link as a directory link. The uuid it carries then "lives in
a non-README file", which reads exactly like a genuine path/uuid conflict and gets reported as one:
`Kind.CONFLICT`, exit 2. Unlike a real conflict, `--write` could never heal it either — the
diagnosis was both false and permanent.

Fix is one `.strip()`, at the single shared primitive all three affected call sites go through
(`names_md`, used by `repair.py`'s `dir_link` check, `robustify.py`'s `_anchor_target`, and
`_dir_link_missing_readme`) — **not** a general whitespace-stripping rule for href resolution, which
the issue itself re-scoped out on 2026-08-13 (that's #74's job; on `main` today no axis strips, so
there's no cross-axis divergence to fix, only this one false classification).

`resolve_href` stays untouched, so a link whose destination genuinely differs by that space still
fails to resolve and gets corrected by the ordinary repair-a-stale-link path — meaning the fix's
side effect is to clean up the stray space when it heals the link, not just stop misdiagnosing it.

Two tests added to `test_conflict.py` (the file that already exists for exactly this finding kind):
the false conflict now repairs and drops the space, and a genuine directory-vs-non-README conflict
(no trailing space) still flags — guarding against overcorrection.

**407 tests pass** (were 405; +2). darnlang clean, repair/robustify/dangling gates clean.

### DIFF
--- a/src/darnlink/paths.py
+++ b/src/darnlink/paths.py
@@ -90,6 +90,16 @@ def names_md(href: str) -> bool:
     uses the href verbatim, so a link whose destination genuinely differs only by that space still
     fails to resolve to the real file and gets corrected by the normal repair-a-stale-link path, not
     silently treated as already-correct.
+
+    ⚠️ `robustify._anchor_target` and `_dir_link_missing_readme` have the SAME "type says file/dir,
+    raw resolution finds nothing" combination for a plain (not-yet-anchored) link -- confirmed by
+    reproduction, not fixed here. #67's own body names both sites explicitly and defers them to #74,
+    which is still open: a prior attempt at exactly this kind of edge-strip (nine review rounds, #62)
+    shipped a "121/121 vs cmark" formulation that still regressed two real files in the fleet, because
+    it reasoned about the destination's *edges* and never measured its *interior* against a real
+    corpus. `tests/test_dangling_links.py::test_whitespace_around_a_destination_is_NOT_stripped_here`
+    pins the current (unstripped) behaviour on the sibling `dangling` axis for exactly this reason --
+    do not widen the strip here to those two call sites without going through #74.
     """
     path_part, _ = split_fragment(href)
     return path_part.strip().lower().endswith(".md")
--- a/src/darnlink/paths.py
+++ b/src/darnlink/paths.py
@@ -79,6 +79,17 @@ def is_absolute_local_path(href: str) -> bool:
 
 def names_md(href: str) -> bool:
     """True if the href's path part names a `.md` file (by suffix). Used to tell a link that points
-    at a file (`foo/README.md`) from one that points at a directory (`foo/`), independent of disk."""
+    at a file (`foo/README.md`) from one that points at a directory (`foo/`), independent of disk.
+
+    Strips surrounding whitespace before checking the suffix (FR-066). A trailing space is never
+    part of what a link destination *means* -- CommonMark does not count it as content -- but before
+    this it made `"old/B.md ".endswith(".md")` false, so `repair` misread a plain file link as a
+    *directory* link. The uuid it carries then lives in a non-README file, which reads as "path and
+    uuid disagree" and is reported as an unhealable CONFLICT -- a false, permanent diagnosis of a
+    link that was actually fine. This is the ONLY strip in the resolution path: `resolve_href` still
+    uses the href verbatim, so a link whose destination genuinely differs only by that space still
+    fails to resolve to the real file and gets corrected by the normal repair-a-stale-link path, not
+    silently treated as already-correct.
+    """
     path_part, _ = split_fragment(href)
-    return path_part.lower().endswith(".md")
+    return path_part.strip().lower().endswith(".md")
--- a/tests/test_conflict.py
+++ b/tests/test_conflict.py
@@ -7,6 +7,7 @@ Reproduces a real-world case: a link to an existing `investigation...md` file ca
 mis-pasted uuid belonging to a different README; old behaviour rewrote the path to that README.
 New behaviour: flag it, touch nothing.
 """
+import sys
 from pathlib import Path
 
 from darnlink.frontmatter_index import build_index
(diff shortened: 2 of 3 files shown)


txemi:
## Adversarial review (Copilot unavailable — quota exhausted, reviews="1" but empty; this substitutes it in full)

Reviewed against the worst-failure hypotheses: overcorrection in `names_md`'s shared use across
`repair`/`robustify`/`--create-readme`, coherence with `_dangling_target`, edge cases of `.strip()`,
and the raw-vs-decoded inconsistency noted in #67's own body.

| Round | Areas checked | Findings | Applied |
|---|---|---|---|
| 1 | `names_md` scope (repair.py, robustify.py's `_anchor_target`/`_dir_link_missing_readme`), `.strip()` edge cases (whitespace-only, leading vs trailing, NBSP, percent-encoding), `_dangling_target` coherence, CI (all matrix legs, not just green ones) | **(a)** Reproduced the exact gap the task asked me to hunt: a **plain** link with a trailing space to a real `.md` file is misreported `DANGLING` by `robustify._anchor_target` — same root cause as #67, different axis. Implemented a `.strip()` extension there. **(b)** That fix broke `tests/test_dangling_links.py::test_whitespace_around_a_destination_is_NOT_stripped_here` — a test written specifically to catch a naive strip re-landing in that exact function, after #74 (open, same author) documented nine review rounds on a predecessor PR where an edge-strip formulation measuring 121/121 against `cmark` still regressed two real files in the fleet (interior-whitespace filenames, e.g. `"my mail file.eml"`). #67's own body was re-scoped 2026-08-13 explicitly deferring `_anchor_target`/`_dir_link_missing_readme` to #74. Reverted (b), kept (a) as a documented, reproduced, but deliberately-deferred gap. **(c)** All four Windows CI legs went genuinely red on push (not a Copilot-quota issue — real `pytest` failures, all four Python versions, identical). Root cause: Win32's `GetFullPathNameW` silently strips a *trailing* space from a path component before the filesystem sees it (not a leading one — measured, not assumed), so the new tests' assumptions about "still resolves to nothing / gets rewritten" don't hold on that OS. | **(a)** Reverted — correctly out of scope per #67's own re-scope note; added a sibling pinned test (`test_a_plain_link_with_only_a_trailing_space_is_also_NOT_anchored_here`) documenting the current, deliberate boundary for the exact shape #67 uses, plus a cross-reference to #74 in the `names_md` docstring so the next reviewer doesn't re-derive this the hard way. **(c)** Fixed: split both affected assertions on `sys.platform.startswith("win")`, keeping the one cross-platform guarantee (no CONFLICT, ever) unconditional. Verified against the actual Windows CI runner, not simulated — confirmed red before, green after (took two pushes: the first fix left a second, same-shaped assertion still red). |

**Seeded-defect check** (required before declaring dry): reverted `names_md`'s `.strip()` to the
pre-fix form (`path_part.lower().endswith(".md")`) and reran `tests/test_conflict.py` — caught
immediately by the existing `test_67_a_trailing_space_in_the_path_is_a_repair_not_an_unhealable_conflict`
(CONFLICT reappears where REPAIR is asserted). No missing coverage there.

**Local runs**: 408 tests pass (Linux). `darnlink .` self-check clean (0 conflicts/unresolved).
`darnlink . --robustify --create-frontmatter --no-create-frontmatter-for README.md` (the MAX
self-check from CI) clean. `darnlang check --ext all` clean. `darnlang` on the commit messages: clean
(gate-enforced at commit time).

**CI (this is what actually gates merge, not the local runs above)**: all 26 checks green —
`ci` test matrix (ubuntu × 4 Python + **windows × 4 Python**, the leg this round found and fixed),
`lang`, `smoke` (both OSes), `ps1-syntax`, `lang-pr` title/body, `privacy-gate` (secret-gated, not
reproducible locally by design — covered by CI, not by me).

**Closing**: round 2 found nothing new after the round-1 fixes landed and CI went green across the
full matrix — closing on the deterministic-gate criterion (suite + darnlang + darnlink's own gates,
now confirmed on every CI leg including the one that was actually broken).

Commits this round: `0bc5ded`, `6d29e8d` on `wt/bloque-d1b`.

PR #98, opened 2026-08-25
A link pending on the default branch is not a broken link

txemi:
A blocking link gate had one finding kind covering **two genuinely different states**, and the
consequence is a deadlock: the red blocks the merge that would clear it.

## The measurement

Today, on a private repo of a fleet that uses this gate. A branch added two files; sibling docs link
to them as `blob/master/<path>`. Those URLs 404 until the branch merges:

```
ok 109 | anchor 0 | mismatch 0 | not-found 6 | unverifiable 2454   ->  exit 4  ->  build red
```

**The branch cannot go green before merging, and cannot merge while red.**

| state | what it means | what the gate should do |
|---|---|---|
| will never resolve | a real break | **cut** |
| not there *yet* | resolves on merge | cutting blocks the fix |

`web_not_found` was answering both. No new kind is needed: **`web_unverifiable` already exists** as
the honest "cannot tell from here" bucket, which is exactly what this is.

And `weblinks.py` already argued for the change, thirteen lines above the site it changes:

> *"A false `web_not_found` in a blocking gate is worse than the crash this replaces"*

A link pending on the default branch, called broken, **is** that false positive.

## Two conditions, and the second is the one that makes it safe

1. the path **exists** in the working tree, and
2. the URL points at **our own repo** (`owner/repo` from `origin`)

Path-exists alone would silently downgrade a real break whenever an unrelated repo happens to share
a filename — and shared filenames are the norm, not the exception (`README.md`, `docs/index.md`).
Only the repo we are actually *in* can have its working tree consulted about where a
`blob/<branch>/…` URL will point after a merge. For any other repo it says nothing.

`_github_slug_from_origin` **reuses the hardening** of its owner-only sibling rather than re-deriving
it — the scrubbed git env (a hook exports `GIT_DIR` and would answer about the *wrong* repository,
silently and with a plausible answer) and the ASCII-anchored regex (without `re.ASCII`, `IGNORECASE`
folds U+0131, so a lookalike host resolves as GitHub).

## Tests: the two that matter are the negative ones

| test | what it protects |
|---|---|
| own repo + path present | the deadlock case → `web_unverifiable` |
| own repo + path **absent** | a real 404 is still a real break |
| **another repo + same-named local file** | **must NOT be downgraded** |
| **no origin slug** | the rung is **inert**, behaves exactly as before |

Verified by seeding, not by reading: dropping the own-repo guard makes **exactly those two negative
tests go red**. 482 tests pass with it in place.

> A rung that changed behaviour when it could not identify the repo would be worse than no rung.

## What this does NOT cover

- **It does not find where a file moved to.** That is unchanged and deliberate ("the LLM layer's
  job"); this only stops calling *not yet published* the same thing as *gone*.
- **It says nothing about the tokenless path.** A 404 without a token was already `web_unverifiable`
  for a different reason and is untouched.
- **It cannot help a detached or bare tree** with no `origin` — there the rung is inert by design,
  which is the fourth test.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,38 @@ All notable changes to darnlink are documented here. The format is based on
 [Keep a Changelog](https://keepachangelog.com/), and this project adheres to
 [Semantic Versioning](https://semver.org/).
 
+## 0.25.0
+
+### A link pending on the default branch is not a broken link
+
+A `blob/<default-branch>/<path>` URL to a file that exists in the working tree but has not reached
+the default branch yet was classified `web_not_found` -> `exit 4`. In a blocking gate that is a
+**deadlock**: the red blocks the merge that would make the link resolve. Measured in the field as
+`not-found 6` on a branch that could not go green before merging and could not merge while red.
+
+It now becomes `web_unverifiable` — the kind that already existed for "cannot tell from here" —
+under **five** conditions, every one of which closes a way a permanently-dead link walked through
+an earlier draft of this rung, green:
+
+| condition | what it stops forgiving |
+|---|---|
+| the ref IS the default branch | `blob/<sha>/…`, a tag, a deleted branch — none of which any merge resolves |
+| the slug is ASCII and matches `origin` | lookalike hosts; `casefold()` collapses U+212A to `k` |
+| the path is confined to the tree | `blob/main//etc/hostname` escaped it entirely |
+| the destination is a FILE | a directory is served under `/tree/`, so `/blob/<dir>` 404s forever |
+| the destination is TRACKED | a gitignored artefact never reaches the default branch |
+
+The base of the path join is `rev-parse --show-toplevel`, not the scanned directory: scanning a
+subdirectory is supported, and there `<path>` (repo-relative) compared against the wrong tree.
+Percent-encoded paths are decoded, so the rung is not silently inert on `%2B` and `%20`.
+
+Unknown origin, non-GitHub remote or unknown default branch -> the rung is **inert** and behaviour
+is exactly as before. A rung that changed behaviour when it could not identify the repo would be
+worse than no rung.
+
+⚠️ **Consumers pin by SHA**, so merging this changes nothing for them until each `darnlink-gate.json`
+`ref` is bumped.
+
 ## [Unreleased]
 
 ### Added
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "darnlink"
-version = "0.24.0"
+version = "0.25.0"
 description = "auto-healing Markdown links — anchor links to a UUID: repair moved links, robustify plain ones"
 readme = "README.md"
 requires-python = ">=3.10"
--- a/src/darnlink/cli.py
+++ b/src/darnlink/cli.py
@@ -493,9 +493,13 @@ def _run_web_check_cli(argv: List[str], fetcher=None) -> int:
 
     token = os.environ.get("GITHUB_TOKEN") or None
     web_out_of_root: List[Path] = []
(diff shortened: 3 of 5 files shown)


txemi:
# Verdict · two rounds, two opposite lenses

| Lens | Hypothesis it chased | Outcome |
|---|---|---|
| **can the guard be fooled?** | a permanently-dead link comes out green | **5 leaks, all reproduced** |
| **does the fix fire at all?** | the rung is dead code with the shape of a fix | **refuted** — it fires, and 8 findings |

## The second review refuted my own main worry, and that mattered

It rebuilt the real case end to end — `origin` set, the two files in the tree, linked as
`blob/master/…`, 404 fetcher, token — and drove it through **`_run_web_check_cli`, not the API**:

```
web_unverifiable 2   <- the two pending files   (before: web_not_found)
web_not_found    2   <- one missing, one in another repo  -> exit 4, correct
```

It also closed the three ways it could have been silently inert: **a real worktree** (`.git` is a
file), **a hook environment** with `GIT_DIR` pointing elsewhere — where it confirmed the env scrub is
load-bearing, not cargo-cult — and the **25 real remotes** of the fleet, all of which resolve.

## The five leaks the first review walked through, green

All of them were links that **no merge will ever resolve**:

| leak | what got forgiven |
|---|---|
| **the ref was never looked at** | `blob/<sha>/x`, a tag, a branch deleted years ago |
| **the join base was the scanned dir** | a broken link compared against the wrong tree |
| **`.exists()` accepts a directory** | `/blob/<dir>` 404s forever; GitHub serves those under `/tree/` |
| **…and a gitignored file** | a build artefact never reaches the default branch |
| **the path was joined raw** | `blob/main//etc/hostname` left the repo entirely |
| **`casefold()`** | collapses U+212A→`k`, U+00DF→`ss` — the very lookalikes to reject |

The sharpest one: **this file already carried `_IMMUTABLE_REF_RE`**, 45 lines below, for exactly the
"this is a SHA, not a branch" distinction — and the rung did not consult it.

## And one that was the opposite of a leak

**Percent-encoded paths made the rung inert.** Folder names here carry a timezone offset, which in a
URL is `%2B` — so it did nothing on the commonest shape in the repo. Not a failure: silence.

## What the second review found that neither the code nor I had

**Consumers pin by SHA.** Nine of them sit on one commit. Merging this changed **nothing** for them.
Now `0.24.0 → 0.25.0` with a CHANGELOG entry that says the ref must be bumped per repo.

## One test does not pin what it claimed, and it says so now

Seeding, not reading, is how I found it: remove the containment check and the escape test **still
passes** — `git ls-files` already rejects anything outside the tree. The check stays (two
comparisons, fails closed, and unlike the tracked check it does not need `git` to run) but its
docstring now states what it **measures** instead of taking the credit.

```
each guard seeded ->  F1: 1 red · F3: 3 red · F5: 1 red · F7: 1 red · F4: 0 (redundant, documented)
restored          ->  488 pass
```

## Closing criterion, and what stays uncovered

Closed by **dry with seeding**, not by "nobody found anything".

1. **A file that only lives on an abandoned branch keeps the gate green forever.** Inherent to the
   design; the message no longer promises the merge will happen.
2. **The verdict line still reads `clean of what could be READ — N unverifiable`**, which the recipe
   itself warns can be a false green. For this rung the axis **did** look — it is "saw it and chose
   to forgive", not "could not see". They share a counter, so this dilutes exactly the signal the
   gate documents as dangerous. Not fixed here.
3. **Case-insensitive filesystems** (macOS/Windows) were not measured: `blob/main/Readme.md` against
   `README.md` on disk 404s on GitHub but would pass `.is_file()`. This repo does ship a Windows
   recipe, so it is in scope.
4. **`_own_repo` has no test of its own**; the wiring is covered only end to end.

PR #104, opened 2026-09-02
perf: memoise path resolution inside an explicit run scope

txemi:
## What

`Path.resolve()` is called on the same paths over and over. Measured on a ~3,900-file documentation tree, one `check` run issues **31,470 calls for only 5,767 distinct paths** — 81.7% repeats — and spends **~24% of its wall clock** inside them. The repeats are structural: `robustify` walks the same file list in several passes and each pass re-resolves the same `f`; `resolve_href()` re-resolves the same link targets.

This memoises them **inside an explicit run scope**.

## The design decision worth reviewing

The cache is **opt-in**: `resolved()` is a plain passthrough unless a `resolve_cache()` scope is open, and `main()` opens exactly one.

An earlier draft made it an always-on process-wide dict cleared at the top of `main()`. Adversarial review reproduced the consequence: `plan_robustify`, `plan_repairs`, `resolve_write_scope` and `in_scope` are all public and **none** of them clears, so an embedder calling them twice around a re-pointed symlink got a stale resolution — and through `apply_robustify`, a **wrong uuid written to disk**. `main()` happened to be the only shipped entrypoint, so it was a trap laid rather than sprung, but the failure mode was silent corruption.

Inverted, the worst case for a library consumer is that it is merely **as slow as it was before this PR**, never wrong.

## Soundness

A non-strict `resolve()` (= `realpath`) can only change if a **symlink, a mount, or a directory-vs-symlink component** changes. Creating a file, creating a `README.md` anchor, or rewriting content cannot change it — and content is the only thing darnlink ever writes. Verified rather than assumed: there is no `mkdir`/`rename`/`replace`/`shutil.move`/`symlink_to`/`unlink` anywhere in `src/`.

Relative paths are **never** memoised: their resolution depends on the process cwd, which is not part of the key. darnlink never chdirs, but `--only-from` accepts relative paths (its help advertises piping `git diff --cached --name-only`), so the key would be ambiguous for an embedder that does.

`frontmatter_index.py` is deliberately **not** converted: its `resolve(strict=True)` runs once per file during the walk, raises rather than returning, and has no repeats to save.

## Measurement

Paired and interleaved in one process (arm A neutralises the scope, so both arms run identical code paths), warm-up discarded, n=7:

| | min | p50 |
|---|---|---|
| no cache | 2.960 s | 3.506 s |
| with cache | 2.408 s | 2.580 s |
| | **−18.6 %** | −26.4 % |

**7/7 paired iterations favour the cache.** An independent re-measurement during review, interleaved in *both* orders on a different tree, got −16.9 % and −19.8 % with **18/18** paired iterations favouring it — so the effect is real and order-independent, not a page-cache artefact.

⚠️ A first attempt used a sequential `9×A then 9×B` design and showed **no difference at all**. That was the design, not the machine. Recorded here because anyone re-checking it the same way will see the same null.

## What the tests pin

`547 passed` (4 new). Specifically:

- no memoisation **outside** a scope — the safety that stops the silent-corruption path above;
- a resolution **never outliving** its scope;
- nesting **not** closing the outer scope early;
- `main()` leaving **no scope open** behind it — structural now (leaving the scope drops the entries), so there is no `clear()` for anyone to delete in six months;
- relative paths not memoised, verified by chdir'ing between two calls.

## Review ledger

| Round | Reviewer | Findings | Outcome |
|---|---|---|---|
| 1 | adversarial subsession | 2 IMPORTANTE, 3 MINOR | all applied — the design was inverted because of finding 1 |
| 2 | Copilot (on opening) | — | pending |

Round 1 also **falsified a claim of mine**: the earlier justification for clearing in `main()` said "the test suite rearranges the tree between calls". It does not — pytest hands every test a fresh `tmp_path`, so keys never collide, and suppressing all 55 clears left 543/543 green. The guard was real; its stated reason was not, and nothing held it up.

Areas round 1 attacked and found clean, with evidence: the 26 substitutions (accounted for per-module, chained expressions checked one by one), cache-key equivalence classes (`a//b`, `a/./b`, `Path('')` vs `Path('.')`, `a/../b`), entrypoint enumeration (every shipped entrypoint is a fresh process), and test-suite leakage (0 accidental stale hits across 2,793 memoised resolutions, 82.4% hit rate — which independently confirms the 81.7% repeat figure).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,36 @@ All notable changes to darnlink are documented here. The format is based on
 
 ## [Unreleased]
 
-_Nothing yet._
+### Path resolution is memoised inside an explicit run scope (~-18%)
+
+`Path.resolve()` was called on the same paths over and over: on a ~3,900-file documentation tree one
+`check` run issues **31,470 calls for only 5,767 distinct paths** — 81.7% repeats — and spends ~24%
+of its wall clock inside them. `robustify` walks the same file list in several passes and each pass
+re-resolves the same file; `resolve_href()` re-resolves the same link targets.
+
+Measured paired and interleaved, warm-up discarded: **min -18.6%**, 7/7 paired iterations favouring
+the cache. Independently re-measured twice during review, in both orders and on a different tree:
+-16.9% / -19.8% and -17.6%, 18/18 and 7/7 paired iterations.
+
+**The cache is opt-in, and that is the load-bearing decision.** `paths.resolved()` is a plain
+passthrough unless a `paths.resolve_cache()` scope is open, and `main()` opens exactly one. Two
+earlier drafts got this wrong in ways adversarial review reproduced end to end, both with the same
+consequence — a stale resolution reaching `apply_robustify` and writing the **wrong uuid to disk**:
+
+- an always-on process-wide dict cleared at the top of `main()`, which the public `plan_*`/`apply_*`
+  functions never clear;
+- then a module global saved and restored around the scope, which two **interleaved** scopes corrupt
+  (the second to leave restores the dict the first had already closed), leaving the cache open
+  process-wide with no run alive. Reproduced with a `ThreadPoolExecutor` running `main()` over
+  several roots. It is now a `ContextVar`, which is per-thread and per-task.
+
+Relative paths are never memoised: their resolution depends on the process cwd, which is not part of
+the key. darnlink never chdirs; an embedder might.
+
+### Added
+
+- `darnlink.paths.resolve_cache()` — context manager opening a memoisation scope for `resolved()`.
+- `darnlink.paths.resolved()` — `Path.resolve()`, memoised inside such a scope.
 
 ## [0.26.0] — 2026-08-28
 
--- a/src/darnlink/cli.py
+++ b/src/darnlink/cli.py
@@ -18,8 +18,8 @@ from pathlib import Path
 from typing import List, Optional
 
 from .frontmatter_index import DEFAULT_EXCLUDES, build_index, index_from_contents, scan_tree
-from .repair import apply_repairs, plan_repairs
 from .paths import resolve_cache, resolved
+from .repair import apply_repairs, plan_repairs
 from .report import Finding, Kind
 from .robustify import apply_robustify, plan_robustify
 from .scope import ScopeError, read_paths_from, resolve_write_scope
--- a/src/darnlink/cli.py
+++ b/src/darnlink/cli.py
@@ -17,6 +17,7 @@ import sys
 from pathlib import Path
 from typing import List, Optional
 
+from .paths import resolve_cache
 from .frontmatter_index import DEFAULT_EXCLUDES, build_index, index_from_contents, scan_tree
 from .repair import apply_repairs, plan_repairs
 from .report import Finding, Kind
@@ -748,6 +749,14 @@ def _make_stdio_encoding_safe() -> None:
 
 
 def main(argv: Optional[List[str]] = None) -> int:
+    # `main()` IS the run: it opens the one resolution-cache scope, and leaving it drops every
+    # entry. Anything that reaches the public plan_*/apply_* functions without coming through here
+    # simply does not memoise -- slower, never stale. `tests/test_paths.py` pins both halves.
+    with resolve_cache():
+        return _main(argv)
+
+
+def _main(argv: Optional[List[str]] = None) -> int:
     _make_stdio_encoding_safe()
     raw = list(sys.argv[1:] if argv is None else argv)
     if raw and raw[0] == "check":  # feature 007: report-only gate subcommand
--- a/src/darnlink/cli.py
+++ b/src/darnlink/cli.py
@@ -17,9 +17,9 @@ import sys
 from pathlib import Path
 from typing import List, Optional
 
-from .paths import resolve_cache
 from .frontmatter_index import DEFAULT_EXCLUDES, build_index, index_from_contents, scan_tree
 from .repair import apply_repairs, plan_repairs
+from .paths import resolve_cache, resolved
 from .report import Finding, Kind
 from .robustify import apply_robustify, plan_robustify
 from .scope import ScopeError, read_paths_from, resolve_write_scope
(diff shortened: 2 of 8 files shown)


txemi:
## Review ledger

Copilot could not review this PR — its review came back empty (*"reached their quota limit"*). That counts as `reviews=1` in the API and reads as green from outside, so it was replaced by adversarial subsessions, each briefed to attack rather than confirm.

| Round | Reviewer | Findings | Outcome |
|---|---|---|---|
| 1 | adversarial subsession | 2 IMPORTANT, 3 MINOR | **applied** — the mechanism was inverted to opt-in because of finding 1 |
| 2 | independent adversarial subsession, briefed not to inherit round 1's conclusions | **1 BLOCKING**, 1 IMPORTANT, 7 MINOR | **applied** in `a73e3a4` |
| 3 | adversarial subsession, scoped to `a73e3a4` | running | — |

### What each round changed, and why it needed the next one

**Round 1 — the cache was always on.** `plan_robustify`, `plan_repairs`, `resolve_write_scope` and `in_scope` are public and none of them cleared; an embedder calling them twice around a re-pointed symlink got a stale resolution and, through `apply_robustify`, **a wrong uuid written to disk**. Reproduced end to end. Fix: memoisation became opt-in.

Round 1 also **falsified the justification** for clearing in `main()` — "the test suite rearranges the tree between calls". It does not: pytest hands every test a fresh `tmp_path`. Suppressing all 55 clears left 543/543 green.

**Round 2 — the opt-in scope was thread-unsafe, which broke this PR's own premise.** The scope was a module global saved and restored around the `with`. With two **interleaved** scopes (A enters, B enters, A exits, B exits), B's restore puts back the dict A had already closed and the cache stays open process-wide with no run alive — an always-on cache with no `clear()`, i.e. exactly what round 1 removed, minus the escape valve. Reproduced with a `ThreadPoolExecutor` running `main()` over several roots, and confirmed independently before applying: 7 entries still cached after every run had finished. Fix: `ContextVar`.

Round 2 also showed `test_main_is_the_run_boundary` was **tautological**: it asserted only that no scope was open before and after. Seeding the removal of `with resolve_cache():` from `main()` — the whole optimisation disabled — left **547/547 green**. It now observes the scope open *during* the run; re-seeding that defect makes it fail.

### Findings dismissed with a reason

None. Every finding from both rounds was applied.

### What the rounds attacked and found clean, with evidence

- **The soundness argument**, verified against a widened grep of every filesystem-mutating call (`mkdir`, `makedirs`, `rename`, `replace`, `shutil.*`, `symlink_to`, `unlink`, `tempfile`, `chdir`, …): zero hits in `src/`. The only write path is `write_text_keep_newlines`.
- **The relative-path guard costs nothing on the hot path**: instrumented over three real runs (358 / 16,658 / 795,968 calls) — **zero** relative paths reach `resolved()`.
- **The 26 substitutions**, accounted for per module and read one by one, chained expressions included.
- **`frontmatter_index.py` being left alone**, confirmed numerically: exactly one resolution per file, zero repeats.
- **The measurement**, re-run with a different and more honest arm A (real pre-change code via `git archive`, separate processes, both orders): −17.6%, 7/7 paired. The original −18.6% is corroborated, not inflated.
- **Publication surfaces**: `privacy-gate` and `lang-pr` green on the head commit; no private repo names, hostnames or internal paths in the commits, the body or the CHANGELOG.

### Not reproducible locally

Nothing. Every gate in this repo's CI (`ci`, `lang-pr`, `privacy-gate`) also runs locally via `tools/check.sh`, and both were run.

txemi:
## Verdict — round 3, and closing

**Round 3 changed no mechanism.** It confirmed the `ContextVar` holds under threads, asyncio, exceptions, failing runs, child threads started inside a scope (they do **not** inherit — verified), `copy_context()` and `fork()`. That is the signal to close the loop: the two earlier rounds each changed the design, this one did not.

| Round | Reviewer | Findings | Outcome |
|---|---|---|---|
| 1 | adversarial subsession | 2 IMPORTANT, 3 MINOR | **applied** — mechanism inverted to opt-in |
| 2 | independent subsession, briefed not to inherit round 1 | **1 BLOCKING**, 1 IMPORTANT, 7 MINOR | **applied** in `a73e3a4` — global → `ContextVar` |
| 3 | subsession scoped to `a73e3a4` | 1 IMPORTANT, 3 MINOR | **applied** in `527976b` — mechanism unchanged |

### Round 3's finding, and it is a fair hit

The concurrency test added in round 2 was **probabilistic**. Measured by re-seeding the pre-`ContextVar` behaviour: it went **green over the regression it is named after ~1 run in 7** (7/50 at the committed shape, 26/50 at two workers). It never failed in false, so CI would have stayed quiet while letting the defect back in.

Round 2 replaced a *tautological* test with a *probabilistic* one. Neither is a gate.

It now forces the interleaving with `Event`s instead of hoping the scheduler produces it — A enters, B enters, A leaves, B leaves, which is the exact order a saved-and-restored global corrupts. Re-measured after the change: **20/20 runs catch the seeded defect, 20/20 stay green on the real code.**

### The other three, all applied

- The docstring claimed a cached resolution **cannot** outlive its run. A generator that opens a scope and is abandoned mid-yield keeps it open until collection — unreachable from darnlink, but the claim was absolute and the code is not. Now qualified.
- **Declared what this cache does not cover**: `frontmatter_index`'s `resolve(strict=True)` is ~45% of all raw resolutions on a ~3,900-file tree and is deliberately excluded (it raises, and issues one call per file with no repeats). The previous message read as if the sweep were complete.
- The import ordering the previous commit *claimed* to fix was still not alphabetical. It is now.

### Seeded defects, and which test catches each

| Seeded | Caught by | Suite |
|---|---|---|
| remove `with resolve_cache():` from `main()` | `test_main_opens_the_scope_and_closes_it` | 1 failed |
| back to the global save/restore | `test_interleaved_scopes_do_not_leak` | **20/20** |
| nested scope creates a new dict | `test_nesting_reuses_the_outer_scope…` | 1 failed |
| `reset(token)` never called | nesting + main | 5 failed |
| memoise relative paths too | `test_relative_paths_are_never_memoised` | 1 failed |
| never store in the dict | 2 tests | 2 failed |

### Findings dismissed with a reason

None, across all three rounds. Every finding was applied.

### Closing

`548 passed` · `tools/check.sh` rc=0 · **26/26 CI checks green** on `527976b`, including `privacy-gate` and `lang-pr`. Nothing in this repo's CI is unreproducible locally, and both were run.

Closing criterion met: **round 3 produced no change to the mechanism.**

PR #105, opened 2026-09-15
recipe: read the config through python when there is no python3, or refuse to guess the ref

txemi:
## Problem

On a Windows Jenkins agent, the gate reported every correct directory link, for example `[topic/](topic/) <!-- uuid: <uuid of topic/README.md> -->`, as needing a repair: `[integrity] repair: 167 → FAIL`. The same tree at the same pin is clean on Linux.

## Root cause: the recipe, not darnlink's logic

`recipes/darnlink-gate` reads `darnlink-gate.json` (ref, mode, excludes…) with `python3`. When `python3` is missing, every `read_cfg` silently falls back to its default. The default ref is **darnlink v0.7.0**, which predates directory links. A Windows venv ships `python.exe` and no `python3`, so on that agent the gate ran v0.7.0 at `mode=check` with no excludes.

Reproduced on Linux by taking `python3` off `PATH`: the same 167 repairs. With `python3` present: clean. darnlink's own Windows CI never saw it because it tests the code, not the recipe without `python3`. The directory-link tests this PR started with (`tests/test_directory_links_cli.py`) pass on `windows-latest` too, and they stay as coverage of the CLI path.

## Fix

- **One interpreter, probed by running it:** `python3`, or `python` when that is the Python 3. Windows can also have a `python3` Store stub that is on `PATH` and does not start. Every JSON read uses it.
- **A config file that cannot be read is a can't-gate, not the default ref.** With `darnlink-gate.json` present, no interpreter, and no `DARNLINK_REF` in the environment, the recipe now calls `bail`: skip with a warning, or exit 4 when fail-closed. It no longer runs a version nobody pinned.
- The existing python3-missing test now declares `DARNLINK_REF` through the environment, as the recipe's own comment recommends for CI. What it asserts is unchanged.

## Tests

- Two new tests, three cases:
  - `python` fallback reads the pinned ref;
  - no interpreter, fail-open → skip;
  - no interpreter, fail-closed → exit 4.
- With the recipe from `main`, all three fail.
- Full suite: 553 passed, 1 skipped (the junction case only runs on Windows).

## For consumers

Changing the recipe changes its sha256. A repo that pins `recipe_sha256` has to re-seal it when it bumps to a version that includes this.


txemi:
**Session review, round 1, on `73f62ee`: no blocking findings.** One non-blocking finding and one perception.

| Claim | Verdict | How I checked |
|---|---|---|
| **R1**: with no `python3`, or a `python3` that does not start (Store stub), plus a `python` that is Python 3, the config is read, and every JSON read goes through `$PY` | **Holds** | Read the recipe at the tip. The probe runs each candidate, so a stub that exits non-zero is skipped. The only `python3` occurrences outside comments are two `bail` messages (L633, L649), not calls. `read_cfg`/`read_cfg_len` use `${PY:-python3}`, and the other four readers use `$PY` |
| **R2**: config present, no interpreter, no `DARNLINK_REF` → darnlink never runs; exit 0 with a warning when fail-open, 4 when fail-closed | **Holds** for fail-closed set through `DARNLINK_GATE_FAIL_CLOSED`, which is the CI wiring the recipe documents | The new `bail` (L358) comes before `run()` is defined (L371) and before any `uvx` call. `REF` at L212 is only a string. The parametrised test covers 0 and 4 |
| **R3**: nothing changes where `python3` exists; the existing test only adds `DARNLINK_REF` | **Holds** | With `python3` runnable, `PY=python3` and every call is the same as before. The diff to that test is the one env line |

**Non-blocking.** With `DARNLINK_REF` in the environment and no interpreter, the recipe does not bail (by design). But `mode`, `excludes` and `fail_closed` from the json then fall back to their defaults **silently**, unless each one is also set through the env. The ref is right, but the policy may not be: a json with `mode: max` would run as `check`, which is a weaker gate with no warning. Suggestion: when `[ -f "$cfg" ] && [ -z "$PY" ]`, print one line naming the json keys that were not applied.

**Perception.** The comment at L116-118 still says *"reading the json needs python3"*. It is now "python3 or python (3.x)".

Not verified by me: the Windows-only junction test (it is skipped off Windows; CI ran it on windows-latest).

